PR #9, opened 2022-08-05
build-and-push-assets.yml // Fix push on tag

gmazzap:
## The issue

When the workflow runs on a tag push, it turns out trying to push from a "detached head" status, failing.

That is due to how the "actions/checkout" action works. It seems that by default the action checks out the current "ref". For tags, the current ref is the tag itself, so the action checks out the tag's commit, ending up in a "detached head" status.

## The solution

To fix that, when the workflow is executed because of tag, before pushing we checkout a **new branch** from the "detached head". Then we add, commit, and push to that new branch.
Finally we move the tag, and only at that point we remove the new branch which is not needed anymore.

In theory, we should also add an additional step to push the new branch, conditionally when running on a tag. However, we opted to have 'push.autoSetupRemote' config set to true, that's easier and simpler, because we don't need to add any condition.

---

Bug fix

The workflow fails when extcuted on a tag push, because an attempt of push from a "detached head" status.

The workflow does not fail anymore when executed on a tag push :)

No


Chrico (commented):
as discussed on weekend: LGTM 👍

PR #10, opened 2022-08-11
Add source code syntax check

AlexP11223:
Adds a basic source code syntax check via `php -l` for all `.php` files in repo.
This can detect that e.g. PHP 8 syntax is used in a project that must support PHP 7.

At first I thought that it needs a parameter for excluding dirs like `vendor` (via `-name 'MYDIR' -prune -o`), but actually it is probably not needed if running at the beginning before installing deps.


Chrico:
I'll close this PR in favor for the implementation of #18

PR #29, opened 2023-01-06
Allow customizing package manager

shvlv:

Feature.

It's not possible to select a package manager to install dependencies for QA workflows.

Yarn and npm definitely go in a different ways to provide specific functionality so it will be nice to run QA with the same package manager as locally. And install dependencies from the related `.lock` file if it exists.

Add `PACKAGE_MANAGER` input and related functionality. Essentially code was picked from the `Build and push assets` workflow (without autodetection).

No. Because by default is `npm` as it is now.

**Other information**:

### DIFF
--- a/.github/workflows/static-analysis-js.yml
+++ b/.github/workflows/static-analysis-js.yml
@@ -43,17 +43,10 @@ jobs:
         with:
           node-version: ${{ inputs.NODE_VERSION }}
           registry-url: ${{ inputs.NPM_REGISTRY_DOMAIN }}
-
-      - name: Cache dependencies
-        uses: actions/cache@v3
-        with:
-          path: ~/.deps_cache
-          key: ${{ runner.os }}-node-${{ inputs.NODE_VERSION }}-${{ inputs.PACKAGE_MANAGER }}-${{ hashFiles('**/package-lock.json', '**/npm-shrinkwrap.json', '**/yarn.lock') }}
+          cache: ${{ inputs.PACKAGE_MANAGER }}
 
       - name: Install dependencies
-        run: |
-          ${{ ((inputs.PACKAGE_MANAGER == 'yarn') && 'yarn config set cache-folder ~/.deps_cache') || 'npm config set cache ~/.deps_cache --global' }}
-          ${{ ((inputs.PACKAGE_MANAGER == 'yarn') && 'yarn') || 'npm install' }}
+        run: ${{ ((inputs.PACKAGE_MANAGER == 'yarn') && 'yarn') || 'npm install' }}
 
       - name: Run ESLint
         run: ./node_modules/.bin/eslint ${{ inputs.ESLINT_ARGS }}
--- a/.github/workflows/static-analysis-js.yml
+++ b/.github/workflows/static-analysis-js.yml
@@ -20,7 +20,7 @@ on:
         type: string
       PACKAGE_MANAGER:
         description: Package manager with which the dependencies should be installed (`npm` or yarn`).
-        default: "npm"
+        default: 'npm'
         required: false
         type: string
     secrets:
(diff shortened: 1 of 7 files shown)


.github/workflows/static-analysis-js.yml
```diff
@@ -39,16 +44,16 @@ jobs:
...
 
-      - name: Cache npm dependencies
+      - name: Cache npm/Yarn dependencies
         uses: actions/cache@v3
         with:
-          path: ~/.npm
-          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
-          restore-keys: |
-            ${{ runner.os }}-node-
+          path: ~/.deps_cache
```
tyrann0us, line 50: IMO, it would be better to read the cache directory from the selected package manager instead of hardcoding it to `~/.deps_cache`. This would also allow us to remove the `${{ inputs.PACKAGE_MANAGER }}` part from the `key` config in the next line because npm and Yarn use different folders.

But please see the next comment on the challenges to get/set the cache directory for Yarn.

.github/workflows/static-analysis-js.yml
```diff
@@ -39,16 +44,16 @@ jobs:
...
+          key: ${{ runner.os }}-node-${{ inputs.NODE_VERSION }}-${{ inputs.PACKAGE_MANAGER }}-${{ hashFiles('**/package-lock.json', '**/npm-shrinkwrap.json', '**/yarn.lock') }}
 
-      - name: Install npm dependencies
-        run: npm install
+      - name: Install dependencies
+        run: |
+          ${{ ((inputs.PACKAGE_MANAGER == 'yarn') && 'yarn config set cache-folder ~/.deps_cache') || 'npm config set cache ~/.deps_cache --global' }}
```
tyrann0us, line 55: I know that GitHub Actions still uses Yarn 1 (see e.g. [here](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#:~:text=from%20master%20%3C796213014%3E%29-,Yarn%201.22.19,-Notes%3A)).

However, should this workflow be run on a system where the user manually installed a newer version, `yarn config set cache-folder` won't work because the settings key changed in Yarn >= 2. There, it's called `cacheFolder` ([documentation](https://yarnpkg.com/configuration/yarnrc#cacheFolder)), and the command would therefore be `yarn config set cacheFolder`.

tyrann0us (commented):
Thanks for bringing this to all workflows. I left a couple of comments which would need to be applied to all occurrences should you agree with them.
Thanks!

.github/workflows/static-analysis-js.yml
shvlv: Hmm, yes it's a good suggestion and it follows author guidelines - https://github.com/actions/cache/blob/main/examples.md#get-npm-cache-directory-using-same-shell

But if we are using `@actions/setup-node` maybe we can just use build-in functionality https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data. With this approach, we can remove the dependency step by just
```
cache: ${{ inputs.PACKAGE_MANAGER }}
```

The additional benefit it should support Yarn 1 and 2 out-of-the-box - https://github.com/actions/setup-node/blob/main/src/cache-utils.ts#L68-L77

tyrann0us: Yes, utilizing existing tools for this would be the best approach. Please apply any changes to all workflows that install npm dependencies. Thanks!

tyrann0us (commented):
Thanks for incorporating the changes! While going through `build-and-push-assets.yml`, I realized that this workflow auto-detects the package manager. IMO, this is great because it saves us from having to specify the package manager in every calling workflow. Can you please update this PR to use the auto-detection as well? Thanks!

tyrann0us (approved):
LGTM. Thanks!

Chrico (approved):

PR #32, opened 2023-01-17
Automated release with Semantic Release

dottxado:

Feature

With this PR I would like to add a new workflow that uses Semantic Release. 
It is lightly based on the workflow my team has used for years. 
Semantic release requires a `release.config.js` file in order to configure the plugins that are needed during the release process. This file is created by the workflow, so there is no more need for that file inside the caller repository, unless you are releasing a custom package and you want to provide that file. 

The release.config.js file is generated from a template with basic setups. Refer to the documentation page for more specifications.

No

**Other information**:


luislard (changes requested):

.github/workflows/release-semantic-automated-npm.yml
```diff
@@ -0,0 +1,105 @@
...
+          fi
+
+      - name: Prepare main filename
+        if: ${{ env.HAS_CONFIG == false }}
+        run: |
+          if [ -f style.css ]; then 
+            MAIN_FILENAME='style.css'
+          elif [ -f index.php ]; then 
+            MAIN_FILENAME='index.php'
+          fi
```
luislard, line 57: I will add expand this to:

```bash
if [ -z "$MAIN_FILENAME" ]; then
  echo "MAIN_FILENAME var is empty"
   if [ -f style.css ]; then 
    MAIN_FILENAME='style.css'
  elif [ -f index.php ]; then 
    MAIN_FILENAME='index.php'
  fi
fi
echo "Using MAIN_FILENAME with value \"$MAIN_FILENAME\""
```

luislard (dismissed):

templates/release-semantic-automated-npm/release.config.ejs
```diff
@@ -0,0 +1,24 @@
+module.exports = {
+  "branches": <%- branches %>,
```
tyrann0us, line 2: Allowing branches to differ from a base setup adds a lot of complexity to the workflow. If a package needs a different branch setup, shouldn't it include a custom release.config.js instead?
```suggestion
  "branches": [
    "main",
    "next",
    {
      "name": "beta",
      "prerelease": true
    },
    {
      "name": "alpha",
      "prerelease": true
    }
  ],
```

templates/release-semantic-automated-npm/release.config.ejs
```diff
@@ -0,0 +1,24 @@
...
+    "@semantic-release/changelog",
+    ["@semantic-release/npm", {
+      "tarballDir": "release",
+      "npmPublish": false
+    }],
+    <% if (mainFilename) { %>
+     ["@semantic-release/exec", {
+        "prepareCmd": "sed -i \"s/Version:.*/Version:     ${nextRelease.version}/\" \"<%- mainFilename %>\" ",
+      }],
+    <% } %>
```
tyrann0us, line 15: `mainFilename` can only be `index.php` or `style.css`. So why add all this logic in the workflow to determine the correct file if we could simply run the search-replace command on both files?
```suggestion
    ["@semantic-release/exec", {
      "prepareCmd": "sed -i \"s/Version:.*/Version:     ${nextRelease.version}/\" index.php style.css 2> /dev/null",
    }],
```
The `2> /dev/null` is to silence the error that one of the files doesn't exist.
(untested!)

templates/release-semantic-automated-npm/release.config.ejs
```diff
@@ -0,0 +1,24 @@
...
+    "@semantic-release/github",
+    ["@semantic-release/git", {
+      "assets": <%- filesToCommit %>,
```
tyrann0us, line 18: Why not hardcode this? `@semantic-release/git` should be smart enough to ignore non-existing files.
```suggestion
      "assets": ["CHANGELOG.md", "composer.json", "index.php", "package-lock.json", "package.json", "style.css"],
```

tyrann0us (changes requested):
I'm sorry for the massive delay. I wanted to take time because there are quite some things that can be simplified.

If you agree with the suggested changes in `templates/release-semantic-automated-npm/release.config.ejs` (basically hardcode everything), we could:
* make it a regular JS file
* remove the need to install `ejs`
* remove the `BRANCHES` workflow input and environment variable
* remove the `MAIN_FILENAME` environment variable
* get rid of multiple steps, massively reducing the overall complexity of the workflow:
  * "Prepare main filename"
  * "Prepare list of files to commit"
  * "Prepare list of branches"
  * "Compile the release.config.js template"
  * "Check the presence of the configuration file" (this is unnecessary anyway; why should `release.config.js` be absent if I just checked out this repo?)

.github/workflows/release-semantic-automated-npm.yml
```diff
@@ -0,0 +1,109 @@
+name: Release
```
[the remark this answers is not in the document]
dottxado: Done

.github/workflows/release-semantic-automated-npm.yml
```diff
@@ -0,0 +1,109 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Check presence of release.config.js
+        run: |
+          HAS_CONFIG=false
+          if [ -f "release.config.js" ]; then 
+            echo "Configuration file release.config.js found" 
+            HAS_CONFIG=true
+          fi
```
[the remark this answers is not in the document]
dottxado: Done

.github/workflows/release-semantic-automated-npm.yml
```diff
@@ -0,0 +1,109 @@
...
+
+      - name: Prepare main filename
+        if: ${{ env.HAS_CONFIG == false }}
```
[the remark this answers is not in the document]
dottxado: Cannot be done....don't ask me why, but it doesn't work. And again don't ask me why, it is no longer working like you have reviewed. I'm pretty sure I tested these cases back when I was working on this.

templates/release-semantic-automated-npm/release.config.ejs
dottxado: Done

templates/release-semantic-automated-npm/release.config.ejs
dottxado: Done, but `2> /dev/null` was outputting an exit code != 0. I had to tweak it a little to apply the idea

templates/release-semantic-automated-npm/release.config.ejs
dottxado: Done

.github/workflows/release-semantic-automated-npm.yml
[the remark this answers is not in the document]
tyrann0us: That is very strange. It would make the code easier to read, but that's just how it is.
@Chrico, or do you maybe know why this apparently doesn't work?

.github/workflows/release-semantic-automated-npm.yml
```diff
@@ -0,0 +1,70 @@
...
+        run: |
+          cp ${GITHUB_WORKSPACE}/workflow-repo/templates/release-semantic-automated-npm/release.config.js .
+
+      - name: Check the presence of the configuration file
+        if: ${{ env.HAS_CONFIG == 'false' }}
+        run: |
+          if [ ! -f "release.config.js" ]; then 
+            echo "Configuration file release.config.js does not exist, aborting!" 
+            exit 1
+          fi
```
tyrann0us, line 60: Why do we need this check? We check for a config file in the repository of the caller workflow (https://github.com/inpsyde/reusable-workflows/pull/32/files#diff-dd37b03d2773c8f29786944f4cf0bdecf07d50ba93edb67cfa566b1df852b3fdR35-R39), and if it doesn't exist, we use the one from the reusable workflows (https://github.com/inpsyde/reusable-workflows/pull/32/files#diff-dd37b03d2773c8f29786944f4cf0bdecf07d50ba93edb67cfa566b1df852b3fdR41-R52).

IMO, the only helpful case for such an additional check would be in an E2E test, but we don't have that yet.
```suggestion
```

docs/release-semantic-automated-npm.md
```diff
@@ -0,0 +1,62 @@
...
+- package-lock.json
+- package.json
+- composer.json
+
+If you don't ship any of these files the release process will not fail.
+
+Another standard set of configurations is on the branches: every push to `main` and `next` branches will release a stable version, every push to `alpha` and `beta` branches will release a pre-release version.
+If you would like another setting, please provide your custom release.config.js file.
+
+Packages based on yarn will not be released using this automation.
```
tyrann0us, line 23: I slightly updated the wording. Any chance you can apply the max characters by line, just like the other Markdown files? It should be 100 or 120. You can use PhpStorm for this. Thanks!

```suggestion
# Automatic release

This workflow utilizes [semantic-release](https://github.com/semantic-release/semantic-release) to create a package release.
Note that you must stick to their [commit message convention](https://github.com/semantic-release/semantic-release#commit-message-format) to use it.

You can provide a `release.config.js` file in your repository to create a custom release that uses the following semantic-release plugins:
- git
- npm
- exec

Otherwise, the workflow will create the release with a standard set of configurations, updating the version in the following files:
- `CHANGELOG.md`
- `composer.json`
- `package-lock.json`
- `package.json`
- `style.css` or `index.php`

By default, every push to the `main` and `next` branches will release a stable version, and every push to the `alpha` and `beta` branches will create a pre-release version.
If you would like another setting, please provide your custom release.config.js file.

Packages based on yarn will not be released using this automation.
```

docs/release-semantic-automated-npm.md
```diff
@@ -0,0 +1,62 @@
...
+    uses: inpsyde/reusable-workflows/.github/workflows/release-semantic-automated-npm.yml@main
+```
+
+### Configuration parameters
+
+#### Inputs
+
+| Name                  | Default                                                                                       | Description                                          |
+|-----------------------|-----------------------------------------------------------------------------------------------|------------------------------------------------------|
+| `NODE_VERSION`        | 18                                                                                            | Node version with which the release will be executed |
```
tyrann0us, line 46: Could you please fix the table? If you reformat the file for the line length with PhpStorm (see previous comment), this should be fixed automatically.

.github/workflows/release-semantic-automated-npm.yml
```diff
@@ -0,0 +1,70 @@
+name: Semantic Release
```
tyrann0us, line 1: Sorry, I changed my mind about the workflow name. We should not reference a third-party tool, IMO, because we might it in the future. What about "Automatic release"?
```suggestion
name: Automatic release
```

If you agree, please also update the file names and references accordingly (i.e., `.github/workflows/automatic-release.yml`). Thanks!

tyrann0us (commented):
We are getting very close. Thanks for incorporating all the changes!

~But:~ 
~> Packages based on yarn will not be released using this automation.~ 

~Wait, I was not aware of this! 🤯 Looks like we definitely need to decide on a package manager, and should we decide on yarn, switch to https://github.com/hongaar/semantic-release-yarn.~

Edit: This is not correct.

.github/workflows/release-semantic-automated-npm.yml
dottxado: Removed :+1:

docs/release-semantic-automated-npm.md
dottxado: Done

.github/workflows/release-semantic-automated-npm.yml
dottxado: I like, done :+1:

.github/workflows/release-semantic-automated-npm.yml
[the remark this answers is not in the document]
dottxado: Seems a problem with the evaluation of booleans, that are turned to strings https://github.com/actions/runner/issues/1173
I've done some other tests based on the comments I've found on this issue, but nothing worked.

tyrann0us (approved):
@dottxado, thank you very much for working on this. Never tested this personally but code-wise LGTM.

luislard (approved):

PR #37, opened 2023-02-03
Plugin archive workflow

Biont:
- [ ] The commit message follows our guidelines

feature

The products team is using multiple individual processes for plugin archive generation. There have been attempts to unify this, but the fact remains that these processes are maintained for each plugin separately.

The ability to share a unified workflow across multiple products would mean a substantial improvement in our workflows and maintenance efforts.

Please note that this is strictly an "build" workflow and not a "release" workflow and should not be confused with one. It is *very* common for us to create a `my-plugin-JIRA-123.zip` build off some feature branch in order to pass on an issue to QA. Given that this constitutes ~98/100 builds that we will be doing, you could go as far as to say that is the primary objective. 

To implement it, I have tried to stay close to existing practices and using existing tools where possible
* Environment (PHP/JS) setup follows existing steps used in other  workflows inside this repository
* Asset compilation & downloading of translations can be done via our respective composer plugins. For this the workflow checks if each of these packages is actually used by the project and then executes the respective command.
* For the actual archive generation, I had to pick a new solution and went with the  [`wp dist-archive`](https://developer.wordpress.org/cli/commands/dist-archive/) command for these reasons:
    * It has builtin support for an exclude list (like `rsync --exclude-from`) for reducing clutter from resulting archives
    * Despite not being bundled, it is an official tool of the "WP-CLI" ecosystem, built for this exact purpose. We hopefully won't need to find a replacement anytime soon

No

**Other information**:
Please feel invited and welcome to squash-merge this PR. Next time I will try to setup something I can more easily replicate with `nectos/act`, but for this instance, I just had to do *a LOT* of trial&error commits


.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,219 @@
...
+
+      - name: Add commit hash to plugin header
+        env:
+          GIT_SHA: ${{ steps.git.outputs.sha_short }}
+        run: 'sed -Ei "s/SHA: .*/SHA: ${GIT_SHA}/g" ${{ inputs.PLUGIN_MAIN_FILE }}'
+
+      - name: Set plugin version header
+        env:
+          PACKAGE_VERSION: ${{ github.event.inputs.packageVersion }}
+        run: 'sed -Ei "s/Version: .*/Version: ${PACKAGE_VERSION}/g" ${{ inputs.PLUGIN_MAIN_FILE }}'
```
AlexP11223, line 169: Can we do it like here? Optional and adds the current version if missing in the input (that is if you enter `somebugfix` it becomes `1.2.3-somebugfix`).

https://github.com/woocommerce/woocommerce-paypal-payments/blob/664a53dd36ed70b68745bd21552b0160d7b58462/.github/workflows/package.yml#L32-L38

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,219 @@
...
+      #
+      # Project setup
+      #
+
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Create plugin basename from specified main file
+        id: basename
+        run: echo "basename=$(basename ${{ inputs.PLUGIN_MAIN_FILE }} .php)" >> $GITHUB_OUTPUT
```
AlexP11223, line 76: Would be great to be able to add a prefix. Like `JIRA-123`.

https://github.com/woocommerce/woocommerce-paypal-payments/blob/664a53dd36ed70b68745bd21552b0160d7b58462/.github/workflows/package.yml#L46-L50

Biont: Instead of relying on the name of the plugin main file, there is now a new input called `PLUGIN_BASENAME`.
If this is not passed, the workflow will fall back to the name of the repository.

.github/workflows/build-plugin-archive.yml
Biont: We discussed this and agreed that the nature of the version string is best defined and maintained by the plugin developers. The reusable workflow would just take the input and use it as-is without attaching any logic or meaning to it.

I added a note about this to the documentation

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,241 @@
...
+        description: Path to a custom php script to run before creating the release archive.
+        default: ''
+        required: false
+        type: string
+      EXTRA_PHP:
+        description: Run custom php code before creating the release archive.
+        default: ''
+        required: false
+        type: string
+    secrets:
```
tyrann0us, line 55: Please add the `PACKAGE_MANAGER` input as seen e.g. here: https://github.com/inpsyde/reusable-workflows/blob/f362c32c9d14f1c27f1586bb1a13a44b5a2e3ab4/.github/workflows/static-analysis-js.yml#L21C7-L25

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,241 @@
...
+          registry-url: ${{ inputs.NPM_REGISTRY_DOMAIN }}
+
+      - name: Cache npm dependencies
+        if: steps.check-for-package.outputs.has_package == 'true'
+        uses: actions/cache@v3
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
```
tyrann0us, line 177: No need for the `actions/cache` package.
```suggestion
          cache: ${{ inputs.PACKAGE_MANAGER }}
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,241 @@
...
+
+      #
+      # Package metadata
+      #
+
```
tyrann0us, line 182:
```suggestion
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,241 @@
...
+      - name: Set plugin version header
+        env:
+          PACKAGE_VERSION: ${{ github.event.inputs.packageVersion }}
+        run: 'sed -Ei "s/Version: .*/Version: ${PACKAGE_VERSION}/g" ${{ inputs.PLUGIN_MAIN_FILE }}'
+
+      #
+      # Archive tooling
+      #
+
```
tyrann0us, line 196:
```suggestion
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,241 @@
...
+      - name: Install WP-CLI
+        run: |
+          mkdir ~/wp-cli && cd ~/wp-cli
+          curl -L https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar --output wp
+          chmod +x wp
+          sudo ln -s ~/wp-cli/wp /usr/local/bin/wp
+
```
tyrann0us, line 203: WP-CLI should be installed with `shivammathur/setup-php`; see my comment above.
```suggestion
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,241 @@
...
+          ${{ inputs.EXTRA_PHP }}
+          EOF
+          )
+          if [ ! -z "$FILE" ]; then
+            PARAMS+="--require=$FILE "
+          fi
+          if [ ! -z "$CODE" ]; then
+            PARAMS+="--exec=${CODE@Q} "
+          fi
+          echo "params=${PARAMS}" >> $GITHUB_OUTPUT
```
tyrann0us, line 222: TBH, I don't understand why we can't utilize GitHub Actions for running PHP commands and need to use the overly complex setup to build `--require`/`--exec` flags for WP-CLI.
```suggestion
      - name: Execute custom code before archive creation
        shell: php {0}
        run: |
          ${{ inputs.EXTRA_PHP_FILE }}
```
(untested!)
Documentation: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#custom-shell
Example: https://github.com/marketplace/actions/setup-php-action#inline-php-scripts

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,241 @@
...
+
+      #
+      # Archive/Artifact creation
+      #
+
```
tyrann0us, line 227:
```suggestion
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,241 @@
...
+      - name: Run wp-cli command
+        run: wp dist-archive . ./archive.zip ${{ steps.dist-archive.outputs.params }} --plugin-dirname=${{ steps.plugin-data.outputs.plugin-name }}
```
tyrann0us, line 229:
```suggestion
      - name: Create archive
        run: wp dist-archive . ./archive.zip --plugin-dirname=${{ steps.plugin-data.outputs.plugin-name }}
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,241 @@
...
+
+        # GitHub Action artifacts would otherwise produce a zip within a zip
+      - name: Unzip archive to dist/
+        run: unzip archive.zip -d dist
+
+      - id: set-artifact-name
+        run: echo "artifact=${{ steps.plugin-data.outputs.plugin-name }}-${{ inputs.PLUGIN_VERSION }}" >> $GITHUB_OUTPUT
+
+      - uses: actions/upload-artifact@v3
```
tyrann0us, line 238: You used named commands previously, so let's continue this here.
```suggestion
      - name: Set artifact name
        id: set-artifact-name
        run: echo "artifact=${{ steps.plugin-data.outputs.plugin-name }}-${{ inputs.PLUGIN_VERSION }}" >> $GITHUB_OUTPUT


      - name: Upload artifact
        uses: actions/upload-artifact@v3
```

docs/archive-creation.md
```diff
@@ -0,0 +1,69 @@
+# Create plugin archive
+
+This action can be used to create plugin archives in a controlled and isolated environment via GitHub Actions.
+
+To achieve that, the reusable workflow:
+
+1. Installs dependencies (including dev-dependencies) defined in `composer.json`
+2. Executes `inpsyde/composer-assets-compiler` if required & configured by the package
+3. Executes `inpsyde/wp-translation-downloader` if required & configured by the package
+4. Re-installs composer dependencies without dev-dependencies
```
tyrann0us, line 10:
```suggestion
4. Re-installs dependencies without dev-dependencies
```

docs/archive-creation.md
```diff
@@ -0,0 +1,69 @@
...
+      EXTRA_PHP: |
+        echo 'hello world!';
+
+```
+
+## Configuration parameters
+
+### Inputs
+
+| Name                  | Default                                                       | Description                                                                    |
```
tyrann0us, line 44: The table Markdown syntax is malformed. Also, we don't include the full stops (`.`) in the description (applies to the other table as well).

docs/archive-creation.md
```diff
@@ -0,0 +1,69 @@
...
+| `NPM_REGISTRY_DOMAIN` | `"https://npm.pkg.github.com/"`                               | Domain of the private npm registry.                                             |
+| `EXTRA_PHP_FILE`      | `""`                                                          | Path to a custom php script to run before creating the release archive.        |
+| `EXTRA_PHP`           | `""`                                                          | Run custom php code before creating the release archive.                       |
+
+#### A note on `PLUGIN_VERSION`
+
+The workflow will accept any arbitrary string and will use it without validation or sanitization. 
+Adding this would mean reduced flexibility at increased complexity. Adding some sort of standardization/validation
+is encouraged, but should take place in the controlling workflow, where the conventions and requirements of the
+project/team/client are known.
```
tyrann0us, line 61:
```suggestion
The workflow will accept any arbitrary string and will use it without validation or sanitization.
Adding this would mean reduced flexibility at increased complexity. Adding standardization/validation
is encouraged but should take place in the controlling workflow, where the conventions and requirements
of the project/team/client are known.
```

tyrann0us (changes requested):
I'm sorry for the massive delay. I wanted to take time because there are quite some things that can be improved.

Apart from the line comments, please also update the documentation accordingly. Thanks!

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,241 @@
...
+      EXTRA_PHP_FILE:
+        description: Path to a custom php script to run before creating the release archive.
+        default: ''
+        required: false
+        type: string
+      EXTRA_PHP:
+        description: Run custom php code before creating the release archive.
+        default: ''
+        required: false
+        type: string
```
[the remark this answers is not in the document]
Biont: I found it neat. First, these just map directly to what's supported by `wp dist-archive`, so I didn't go out of my way to make it happen.

Second, the beuaty of both is that
* `EXTRA_PHP_FILE` can be used for somewhat more complex scripts that need their own place in the repository (and might be reused by other components)
* `EXTRA_PHP` can be inlined directly into the workflow YAML for one-shot adaptions, see [for example here](https://github.com/inpsyde/ddev-wordpress-plugin-template/blob/reusable-release-workflow/.github/workflows/release.yml#L17-L18).

I mean this is not going to be the hill I'm going to die on, but I do believe both options have their merit

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,241 @@
...
+    steps:
+
+      #
+      # Project setup
+      #
+
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup package base name
```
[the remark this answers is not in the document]
Biont: Yeah. Initially I wanted to pretend this could work for themes as well, which is why I used "package".

I've since dropped that idea, but didn't update the wording everywhere

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,241 @@
...
+        id: plugin-data
+        run: |
+          PLUGIN_BASENAME='${{ inputs.PLUGIN_BASENAME }}'
```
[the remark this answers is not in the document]
Biont: Seems inconsequential, but OTOH, it's less shell, more YAML. So probably a good change. Thanks.
No idea if this can introduce subtle changes in behaviour, but there's only one way to find out. We'll see

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,241 @@
...
+              .forEach(envVar => core.exportVariable(envVar.name, envVar.value));
+
+      #
+      # PHP dependencies
+      #
+
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ inputs.PHP_VERSION }}
```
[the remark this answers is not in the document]
Biont: Hooray, I didn't know this was possible. Thanks

.github/workflows/build-plugin-archive.yml
Biont: Cool!

.github/workflows/build-plugin-archive.yml
Biont: The difference is that this executes naked PHP without any context whatsoever.
If passed to WP-CLI instead, you execute within its [bootstrapping process](https://github.com/wp-cli/wp-cli/blob/2efbaad2d5def76c5fb78b758f98f3b9c734e694/php/bootstrap.php#L28-L29), giving you access to its autoloader, shell utils and possibly also limited access to its execution context.

Granted: It unfortunately does not provide you with a bootstrapped WordPress (->access to core functions), but I'd say this is still miles better than _raw PHP_  (at that point, I'd prefer bash/sh for its vastly more powerful file/directory modification tools).

.github/workflows/build-plugin-archive.yml
Biont: Added

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,183 @@
...
+        description: The new plugin version.
+        required: true
+        type: string
+      ARCHIVE_NAME:
+        description: The base name of the resulting zip archive. Falls back to the repository name.
+        default: ''
+        required: false
+        type: string
+      COMPOSER_ARGS:
+        description: Set of arguments passed to Composer for production setup.
```
tyrann0us, line 21: Now that we no longer have `COMPOSER_ARGS_DEV`, please adjust this description accordingly. 
```suggestion
        description: Set of arguments passed to Composer.
```

docs/archive-creation.md
```diff
@@ -0,0 +1,68 @@
...
+## Configuration parameters
+
+### Inputs
+
+| Name                  | Default                                                       | Description                                                                        |
+|-----------------------|---------------------------------------------------------------|------------------------------------------------------------------------------------|
+| `PLUGIN_MAIN_FILE`    | `"index.php"`                                                 | The name of the plugin main file.                                                  |
+| `PLUGIN_VERSION`      | -                                                             | The plugin version.                                                                |
+| `ARCHIVE_NAME`        | `""`                                                          | The base name of the resulting zip archive. Falls back to the repository name.     |
+| `COMPOSER_ARGS`       | `'--no-dev --no-scripts --prefer-dist --optimize-autoloader'` | Set of arguments passed to Composer for production setup.                          |
```
tyrann0us, line 48:
```suggestion
| `COMPOSER_ARGS`       | `'--no-dev --no-scripts --prefer-dist --optimize-autoloader'` | Set of arguments passed to Composer.                                               |
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,183 @@
...
+            JSON
+              .parse(process.env.ENV_VARS)
+              .forEach(envVar => core.exportVariable(envVar.name, envVar.value));
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ inputs.PHP_VERSION }}
+          tools: wp-cli
+
+      - name: Install Composer dependencies with dev dependencies
```
tyrann0us, line 105: To align with other workflows:
```suggestion
      - name: Install Composer dependencies
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,183 @@
...
+        # ⬑ The idea is to ensure we have non-production tools like composer-assets-compiler and translation downloader available
```
tyrann0us, line 106: Let's use the "official" names in the comment.
```suggestion
        # ⬑ The idea is to ensure we have non-production tools like Composer Asset Compiler and WordPress Translation Downloader available
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,183 @@
...
+        run: |
+          hasDep(){
+            composer show -i -N -D --strict "${1}" >/dev/null 2>&1
+            local EXIT=$?
+            echo "$EXIT"
+          }
+          echo "assets-compiler=$( hasDep inpsyde/composer-assets-compiler )" >> $GITHUB_OUTPUT
+          echo "translation-downloader=$( hasDep inpsyde/wp-translation-downloader )" >> $GITHUB_OUTPUT
+
+      - name: Run Composer Assets Compiler
```
tyrann0us, line 122: The official name is "Composer Asset Compiler", without the "s".
```suggestion
      - name: Run Composer Asset Compiler
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,183 @@
...
+      - name: Run WP Translation Downloader
+        if: steps.composer-tools.outputs.translation-downloader == '0'
+        run: composer wp-translation-downloader:download
+
+      - name: Install Composer dependencies without dev dependencies
+        # ⬑ Now we want to get rid of dev dependencies and most probably skip composer scripts
+        uses: ramsey/composer-install@v2
+        with:
+          composer-options: ${{ inputs.COMPOSER_ARGS }}
+      - name: Check for package.json
```
tyrann0us, line 135:
```suggestion

      - name: Check for package.json
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,183 @@
...
+        env:
+          ENV_VARS: ${{ secrets.ENV_VARS }}
+        if: ${{ env.ENV_VARS }}
+        uses: actions/github-script@v6
+        with:
+          script: |
+            JSON
+              .parse(process.env.ENV_VARS)
+              .forEach(envVar => core.exportVariable(envVar.name, envVar.value));
+      - name: Set up PHP
```
tyrann0us, line 99:
```suggestion

      - name: Set up PHP
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -0,0 +1,183 @@
...
+            PLUGIN_VERSION: ${{ inputs.PLUGIN_VERSION }}
+        run: 'sed -Ei "s/Version: .*/Version: ${PLUGIN_VERSION}/g" ${{ inputs.PLUGIN_MAIN_FILE }}'
+      - name: Install dist-archive command
+        run: wp package install wp-cli/dist-archive-command
+
+      - name: Execute custom code before archive creation
+        run: |
+            ${{ inputs.EXTRA_PHP_FILE }}
+
+      - name: Run wp-cli command
```
tyrann0us, line 168: The official name is "WP-CLI".
```suggestion
      - name: Run WP-CLI command
```

docs/archive-creation.md
tyrann0us, line 1: As stated [before](https://github.com/inpsyde/reusable-workflows/pull/37#discussion_r1182329928), please remove the full stop (`.`) from the description column.

tyrann0us (approved):
Thanks for incorporating the changes! LGTM.

PR #39, opened 2023-02-15
Make possible to specify a working directory

widoz:
There might be edge cases where a project might have different modules or packages requiring a specific configuration to build the assets. In most cases the problem is solvable with monorepo strategies but it's not always possible due to a different set of reasons, mainly legacy reasons and different unresolvable dependencies hierarchies.

Github allow to specify a `workflow-directory` different than the project root which would help in such cases.

The changes are non intrusives and are an opt-in enhancement, hence they will not force any consumer to adapt to the new configuration.

### DIFF
--- a/.github/workflows/build-and-push-assets.yml
+++ b/.github/workflows/build-and-push-assets.yml
@@ -4,7 +4,7 @@ on:
   workflow_call:
     inputs:
       WORKING_DIRECTORY:
-        description: The Working Directory.
+        description: Working directory path.
         default: './'
         required: false
         type: string
--- a/.github/workflows/build-and-push-assets.yml
+++ b/.github/workflows/build-and-push-assets.yml
@@ -4,7 +4,7 @@ on:
   workflow_call:
     inputs:
       WORKING_DIRECTORY:
-        description: Set a different working directory for the job.
+        description: The Working Directory.
         default: './'
         required: false
         type: string
--- a/.github/workflows/build-and-push-assets.yml
+++ b/.github/workflows/build-and-push-assets.yml
@@ -3,6 +3,11 @@ name: Build and push assets
 on:
   workflow_call:
     inputs:
+      WORKING_DIRECTORY:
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-push-assets.yml
```diff
@@ -3,6 +3,11 @@ name: Build and push assets
 on:
   workflow_call:
     inputs:
+      WORKING_DIRECTORY:
+        description: The Working Directory.
```
tyrann0us, line 7:
```suggestion
        description: Working directory path.
```

tyrann0us (commented):
Thanks for this! Please allow me two comments:
1. Do you think having the work dir configurable in all available workflows makes sense? If so, it would be really great if you could add it everywhere.
2. Please update the documentation.

Thanks!

gmazzap:
@tyrann0us looking at other workflows, I can't think of a reason why it would be necessary to change the working dir.

I've committed your suggestion and updated the docs.

If we can merge this morning, it would be really helpful because it is a blocker for us. Otherwise, we'll need to require the PR's source branch to unblock our work.

gmazzap (approved):

tyrann0us (approved):
> If we can merge this morning, it would be really helpful because it is a blocker for us.

Understood. Merging now.

widoz:
@gmazzap @tyrann0us Thanks a bunch for the corrections.

> Do you think having the work dir configurable in all available workflows makes sense? If so, it would be really great if you could add it everywhere.

Honestly I don't know, for the assets I had a real case scenario but unfortunately I do not use much the reusable workflows yet and I don't know. For sure it would be a flexibility that does not produce side effects, so why not? 🤷🏽 

That being said, for PHPCS for instance I found difficult to have a different working directory and probably it highly depends on how we organize our code. 

In conclusion if it would be necessary we can still add it, we know this feature exists.

Right now I really need this so I would merge it.

PR #54, opened 2023-07-25
Allow installing packages with SSH in all node-workflows

shvlv:

Feature.

We can't install packages with a `git+ssh` URL.

The approach is adopted from assets compilation workflows to configure the SSH agent. This way we can install packages from GitHub repositories directly.

No.

**Other information**:

### DIFF
--- a/.github/workflows/wp-scripts-lint.yml
+++ b/.github/workflows/wp-scripts-lint.yml
@@ -100,7 +100,7 @@ jobs:
         with:
           ssh-private-key: ${{ secrets.GITHUB_USER_SSH_KEY }}
 
-      - name: Setup Git
+      - name: Set up Git
         if: ${{ env.GITHUB_USER_EMAIL != '' && env.GITHUB_USER_NAME != '' }}
         run: |
           git config --global user.email "${{ secrets.GITHUB_USER_EMAIL }}" 
--- a/.github/workflows/wp-scripts-lint.yml
+++ b/.github/workflows/wp-scripts-lint.yml
@@ -58,10 +58,10 @@ on:
         required: false
       GITHUB_USER_EMAIL:
         description: Email address for the GitHub user configuration.
-        required: true
+        required: false
       GITHUB_USER_NAME:
         description: Username for the GitHub user configuration.
-        required: true
+        required: false
       GITHUB_USER_SSH_KEY:
         description: Private SSH key associated with the GitHub user passed as `GITHUB_USER_NAME`.
         required: false
@@ -76,9 +76,9 @@ jobs:
     env:
       NODE_OPTIONS: ${{ inputs.NODE_OPTIONS }}
       NODE_AUTH_TOKEN: ${{ secrets.NPM_REGISTRY_TOKEN }}
-      GITHUB_USER_SSH_KEY: ${{ secrets.GITHUB_USER_SSH_KEY }}
-      GITHUB_USER_NAME: ${{ secrets.GITHUB_USER_NAME }}
       GITHUB_USER_EMAIL: ${{ secrets.GITHUB_USER_EMAIL }}
+      GITHUB_USER_NAME: ${{ secrets.GITHUB_USER_NAME }}
+      GITHUB_USER_SSH_KEY: ${{ secrets.GITHUB_USER_SSH_KEY }}
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


.github/workflows/wp-scripts-lint.yml
```diff
@@ -67,6 +76,9 @@ jobs:
     env:
       NODE_OPTIONS: ${{ inputs.NODE_OPTIONS }}
       NODE_AUTH_TOKEN: ${{ secrets.NPM_REGISTRY_TOKEN }}
+      GITHUB_USER_SSH_KEY: ${{ secrets.GITHUB_USER_SSH_KEY }}
+      GITHUB_USER_NAME: ${{ secrets.GITHUB_USER_NAME }}
+      GITHUB_USER_EMAIL: ${{ secrets.GITHUB_USER_EMAIL }}
```
tyrann0us, line 81: Why do we set these environment variables here if we read the values from the _secrets_ later and only use the env vars to check for existence (which is unnecessary in one of two cases)?
```suggestion
```
If you agree with this change, please also update accordingly in https://github.com/inpsyde/reusable-workflows/blob/main/.github/workflows/build-and-push-assets.yml. Thanks!

.github/workflows/wp-scripts-lint.yml
```diff
@@ -82,6 +94,18 @@ jobs:
               .parse(process.env.ENV_VARS)
               .forEach(envVar => core.exportVariable(envVar.name, envVar.value));
 
+      - name: Set up SSH
+        if: ${{ env.GITHUB_USER_SSH_KEY != '' }}
+        uses: webfactory/ssh-agent@v0.8.0
+        with:
+          ssh-private-key: ${{ secrets.GITHUB_USER_SSH_KEY }}
+
+      - name: Setup Git
```
tyrann0us, line 103:
```suggestion
      - name: Set up Git
```
Could you please fix this here as well, according to the pathfinder rules?
https://github.com/inpsyde/reusable-workflows/blob/f0fbb4ebdb4628bb40f9a949c3646720875e0e6a/.github/workflows/build-and-push-assets.yml#L116
Thanks!

.github/workflows/wp-scripts-lint.yml
```diff
@@ -82,6 +94,18 @@ jobs:
...
+        if: ${{ env.GITHUB_USER_EMAIL != '' && env.GITHUB_USER_NAME != '' }}
```
tyrann0us, line 104: `GITHUB_USER_EMAIL` and `GITHUB_USER_NAME` are required, so there is no need to check them here again.
```suggestion
```

.github/workflows/wp-scripts-lint.yml
```diff
@@ -82,6 +94,18 @@ jobs:
               .parse(process.env.ENV_VARS)
               .forEach(envVar => core.exportVariable(envVar.name, envVar.value));
 
+      - name: Set up SSH
+        if: ${{ env.GITHUB_USER_SSH_KEY != '' }}
```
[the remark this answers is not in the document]
shvlv: It does not work https://docs.github.com/en/actions/security-guides/encrypted-secrets#:~:text=Secrets%20cannot%20be%20directly%20referenced%20in%20if%3A%20conditionals.%20Instead%2C%20consider%20setting%20secrets%20as%20job%2Dlevel%20environment%20variables%2C%20then%20referencing%20the%20environment%20variables%20to%20conditionally%20run%20steps%20in%20the%20job.

.github/workflows/wp-scripts-lint.yml
shvlv: yes, I copied most of the code from https://github.com/inpsyde/reusable-workflows/blob/f0fbb4ebdb4628bb40f9a949c3646720875e0e6a/.github/workflows/build-assets-compilation.yml. But secret definitions from https://github.com/inpsyde/reusable-workflows/blob/f0fbb4ebdb4628bb40f9a949c3646720875e0e6a/.github/workflows/build-and-push-assets.yml#L55-L63 :smile: . I will update PR to make all new secrets optional.

tyrann0us (commented):
Thanks for the feedback. I know this changes the original scope of the PR, but as a pathfinder rule and to align all workflows, could you please 
1. add env variables for `GITHUB_USER_EMAIL` and `GITHUB_USER_NAME` in `build-and-push-assets.yml`
2. read `GITHUB_USER_*` from the respective env variables in `build-and-push-assets.yml` and `build-assets-compilation.yml` in the "Set up SSH" and "Set up Git" steps
3. sort the declaration of the `GITHUB_USER_*` env variables to follow the declaration of the secrets in all workflows 🙈  (i.e., `GITHUB_USER_EMAIL`, `GITHUB_USER_NAME`, `GITHUB_USER_SSH_KEY`)
4. Add `GITHUB_USER_*` secrets to _all_ remaining workflows (and documentations)

Thanks a ton!

Edit: Added point 4 after a discussion with @Chrico.

tyrann0us (approved):
LGTM. Thanks!

PR #60, opened 2023-08-14
Introduce Dry Run option

widoz:
Allow third parties to perform a compilation without committing or pushing the assets.

This is useful in a CI context when in a PR you need to compile the assets to prevent wrong changes to affect the target branch.

An example for a possible configuration is `DRY_RUN: ${{ github.event_name != 'pull_request' }}`

---

Let's assume your colleague made a PR, the PR includes changes in some .ts|.tsx files. The action is not running in the PR but only in the target branch and only after the merge. 

It might happen that after the merge the compilation of the assets failed, now you have a broken build. 

To prevent that, we would like to run the assets compilation in the PR but the branch associated to the PR does not need to get the assets updated since them will be compiled when the PR will be merged. 

Having the dry-run option ON will allow us to run the compilation as part of the CI of the PR to ensure the changes are not impacting negatively the project. Nonetheless we do not want to commit and push the assets in the branch but only when the PR is merged, hence in the target branch.

It is the target branch responsible to commit and push the assets to avoid conflicts and problems.

### DIFF
--- a/.github/workflows/build-and-push-assets.yml
+++ b/.github/workflows/build-and-push-assets.yml
@@ -48,6 +48,11 @@ on:
         default: './assets'
         required: false
         type: string
+      DRY_RUN:
+        description: Compile the assets but do not push them to the repository.
+        default: false
+        required: false
+        type: boolean
     secrets:
       NPM_REGISTRY_TOKEN:
         description: Authentication for the private npm registry.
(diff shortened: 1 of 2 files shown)


widoz:
Closing in favor of https://github.com/inpsyde/reusable-workflows/issues/63

PR #65, opened 2023-08-18
Fix stylelint arguments passing

shvlv:

Bug fix.

Stylelinting doesn't work as expected if we follow the documentation.
If the lint-style command uses a glob pattern it must be quoted - https://developer.wordpress.org/block-editor/reference-guides/packages/packages-scripts/#lint-style and https://stylelint.io/user-guide/cli/#:~:text=stylelint%20%22**/*.css%22-,You%20should%20include%20quotation%20marks%20around%20file%20globs.,-If%20you%20are.

Quote arguments (the same as here https://github.com/inpsyde/reusable-workflows/commit/f362c32c9d14f1c27f1586bb1a13a44b5a2e3ab4).

No.

**Other information**:
Alternatively, we could improve the documentation to force provide quoted arguments: `STYLELINT_ARGS: '"./resources/**/*.scss"'` but I'm not sure it's a good idea because users will forget about this.

### DIFF
--- a/.github/workflows/wp-scripts-lint.yml
+++ b/.github/workflows/wp-scripts-lint.yml
@@ -131,7 +131,7 @@ jobs:
 
       - name: Lint script files
         if: ${{ contains(fromJSON(inputs.LINT_TOOLS), 'js') }}
-        run: ./node_modules/.bin/wp-scripts lint-js ${{ inputs.ESLINT_ARGS }}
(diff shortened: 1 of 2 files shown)


tyrann0us (approved):
LGTM. Thank you for noticing! Does it make sense to quote the other lint tool arguments, too?

shvlv:
> LGTM. Thank you for noticing! Does it make sense to quote the other lint tool arguments, too?

It looks like other tools work as expected while it will not be harm :man_shrugging:

PR #79, opened 2023-10-23
feat: adds possibility to install WP_CLI and the possibility to use t…

luislard:
…he PROD compile script on specified branches

Adds the ability to configure the environment to use the WP CLI and PHP, these will help if the consumer requires to:

- Use the WP CLI to build the POT files.

We are also adding the possibility to use the COMPILE_SCRIPT_PROD on the branches workflow since this was not possible.

**What is the current behavior?** (You can also link to an open issue here).

1. There is no possibility of using the WP CLI.
2. There is no possibility to use the COMPILE_SCRIPT_PROD in the branches

Consumers can:

1. Optionally add the WP CLI by configuring it along with the PHP version to use
2. Optionally define an array of branches to run the COMPILE_SCRIPT_PROD

No

**Other information**:

### DIFF
--- a/.github/workflows/build-and-push-assets.yml
+++ b/.github/workflows/build-and-push-assets.yml
@@ -53,6 +53,21 @@ on:
         type: string
         default: ''
         required: false
+      INSTALL_WP_CLI:
+        description: Whether to Install WP cli or not.
+        default: false
+        required: false
+        type: boolean
+      PHP_VERSION:
+        description: PHP version with which the WP CLI is going to be executed.
+        default: "8.1"
+        required: false
+        type: string
+      BRANCHES_USING_COMPILE_SCRIPT_PROD:
+        description: Branches that should use the COMPILE_SCRIPT_PROD script
+        default: '[]'
+        required: false
+        type: string
     secrets:
       NPM_REGISTRY_TOKEN:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-push-assets.yml
```diff
@@ -53,6 +53,21 @@ on:
...
+      PHP_VERSION:
+        description: PHP version with which the WP CLI is going to be executed.
+        default: "8.1"
+        required: false
+        type: string
+      BRANCHES_USING_COMPILE_SCRIPT_PROD:
+        description: Branches that should use the COMPILE_SCRIPT_PROD script
+        default: '[]'
+        required: false
+        type: string
```
Chrico, line 70: Isn't this something whcih should go into the calling workflow via `on.push.branches` ? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore

tyrann0us (commented):
I left some suggestions for typos. The actual change, I can't really assess, though.

It might make sense to add an input for _any_ PHP tool supported by `shivammathur/setup-php` instead of hardcoding to WP-CLI.
CC @Chrico

PR #87, opened 2023-11-02
Feat/86 make token configurable via new secret GITHUB_USER_TOKEN

### LINKED ISSUE #86: [Feature Request]: Current release reusable workflow does not expose a way of changing the token used or auth method

Current used secret name is too wide, and cannot be overridden.
See [here](https://github.com/inpsyde/reusable-workflows/blob/af929128e7db96c868c54a9b04f18d0bad4885b9/.github/workflows/automatic-release.yml#L60)

I suggest to make it configurable 

Considered using a secret config USER_TOKEN.

Consider using SSH configuration.

### Additional context

_No response_

### Code of Conduct

luislard:
- [ ] The commit message follows our guidelines

Closes #86 

Makes the token configurable so consumers are able to use specific user tokens to for example bypass some branch protection rules.

Token is not configurable

The used token is configurable

No 

**Other information**:

### DIFF
--- a/.github/workflows/automatic-release.yml
+++ b/.github/workflows/automatic-release.yml
@@ -29,6 +29,7 @@ jobs:
         run: |
           npm i -g @semantic-release/changelog \
             @semantic-release/git \
+            @semantic-release/github \
             @semantic-release/npm \
             @semantic-release/exec \
             semantic-release
--- a/.github/workflows/automatic-release.yml
+++ b/.github/workflows/automatic-release.yml
@@ -10,7 +10,7 @@ on:
     secrets:
       USER_TOKEN:
         description: Authentication for the semantic release action.
-        required: true
+        required: false
 
 jobs:
   release:
@@ -64,5 +64,5 @@ jobs:
 
       - name: Release
         env:
-          GITHUB_TOKEN: ${{ secrets.USER_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.USER_TOKEN != '' && secrets.USER_TOKEN || secrets.GITHUB_TOKEN }}
         run: npx semantic-release
--- a/.github/workflows/automatic-release.yml
+++ b/.github/workflows/automatic-release.yml
@@ -35,6 +35,8 @@ jobs:
 
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       - name: Check presence of release.config.js
         run: |
--- a/.github/workflows/automatic-release.yml
+++ b/.github/workflows/automatic-release.yml
@@ -61,5 +61,5 @@ jobs:
 
       - name: Release
         env:
-          GITHUB_TOKEN: ${{ secrets.USER_TOKEN != '' && secrets.USER_TOKEN || secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.USER_TOKEN }}
         run: npx semantic-release
--- a/.github/workflows/automatic-release.yml
+++ b/.github/workflows/automatic-release.yml
@@ -10,7 +10,7 @@ on:
     secrets:
       USER_TOKEN:
         description: Authentication for the semantic release action.
-        required: false
+        required: true
 
 jobs:
   release:
@@ -64,5 +64,5 @@ jobs:
 
       - name: Release
         env:
-          GITHUB_TOKEN: ${{ secrets.USER_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.USER_TOKEN != secrets.USER_TOKEN && secrets.USER_TOKEN || secrets.GITHUB_TOKEN }}
         run: npx semantic-release
--- a/.github/workflows/automatic-release.yml
+++ b/.github/workflows/automatic-release.yml
@@ -9,7 +9,7 @@ on:
         type: string
     secrets:
       USER_TOKEN:
-        description: Authentication for the automatic release action.
+        description: Authentication token with write permission needed by the release bot.
         required: false
 
 jobs:
--- a/.github/workflows/automatic-release.yml
+++ b/.github/workflows/automatic-release.yml
@@ -7,6 +7,10 @@ on:
         default: 18
         required: false
         type: string
+    secrets:
+      GITHUB_TOKEN:
+        description: Authentication for the semantic release action.
+        required: false
 
 jobs:
   release:
--- a/.github/workflows/automatic-release.yml
+++ b/.github/workflows/automatic-release.yml
@@ -8,9 +8,10 @@ on:
         required: false
         type: string
     secrets:
-      GITHUB_TOKEN:
+      USER_TOKEN:
         description: Authentication for the semantic release action.
         required: false
+        default: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
   release:
@@ -61,5 +62,5 @@ jobs:
 
       - name: Release
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.USER_TOKEN }}
         run: npx semantic-release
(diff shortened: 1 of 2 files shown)


.github/workflows/automatic-release.yml
```diff
@@ -7,6 +7,10 @@ on:
         default: 18
         required: false
         type: string
+    secrets:
+      USER_TOKEN:
+        description: Authentication for the semantic release action.
```
dottxado, line 12: During the development of the action we decided to "hide" the semantic release naming under the hood of "automatic release"
```suggestion
        description: Authentication for the automatic release action.
```

dottxado (changes requested):
Hello @luislard ! Please update the [docs here](https://github.com/inpsyde/reusable-workflows/blob/feat/86_make_token_configurable/docs/automatic-release.md) :muscle:

.github/workflows/automatic-release.yml
```diff
@@ -7,6 +7,10 @@ on:
         default: 18
         required: false
         type: string
+    secrets:
+      USER_TOKEN:
+        description: Authentication for the automatic release action.
```
tyrann0us, line 12: Let's be more specific about the token's purpose. It is necessary to allow the `semantic-release-bot` to push to the repository, correct? Then let's write that.
```suggestion
        description: Authentication for write permission for the release bot.
```
I'm not really convinced by the wording, either. Do you have a better suggestion? In any case, please align this string with the one from the documentation.

docs/automatic-release.md
```diff
@@ -58,4 +59,6 @@ jobs:
     uses: inpsyde/reusable-workflows/.github/workflows/automatic-release.yml@main
     with:
       NODE_VERSION: 16
+    secrets:
+      USER_TOKEN: ${{ secrets.MY_TOKEN_SECRET }}
```
tyrann0us, line 63: Let's use a more "professional` token name in the example. 😉
```suggestion
      USER_TOKEN: ${{ secrets.WRITE_TOKEN }}
```

.github/workflows/automatic-release.yml
dottxado: Maybe...?
```suggestion
        description: Authentication token with write permission needed by the release bot.
```

docs/automatic-release.md
Chrico: Is it only a write token or could we also inject a read or read-write token here? does that make any difference? I would suggest to name it just `TOKEN`.

.github/workflows/automatic-release.yml
```diff
@@ -31,6 +35,8 @@ jobs:
 
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false
```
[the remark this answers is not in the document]
luislard: the checkout action also configures git by default, this is a problem because we want to set the user with our token instead, I tested removing it and it does not work. Proof:

Commit removing it: https://github.com/inpsyde/reusable-workflows/pull/87/commits/ef4a166c218b08cb7891138bd1f16e1fad04d371
Corresponding test: https://github.com/inpsyde/mercury-blocks/actions/runs/6744106204/job/18333268117

Here is [the documentation](https://github.com/actions/checkout#:~:text=persist%2Dcredentials%3A%20%27%27) regarding this config. Current default is true.

Also here is an issue / discussion around it https://github.com/actions/checkout/issues/485

docs/automatic-release.md
```diff
@@ -40,9 +40,10 @@ jobs:
 
 #### Inputs
 
-| Name           | Default | Description                                          |
-|----------------|---------|------------------------------------------------------|
-| `NODE_VERSION` | 18      | Node version with which the release will be executed |
+| Name           | Required | Default | Description                                                                                |
+|----------------|----------|------------------|--------------------------------------------------------------------------------------------|
+| `NODE_VERSION` | false    |18               | Node version with which the release will be executed                                       |
+| `USER_TOKEN`   | false    |`empty`          | The token used to authenticate with GitHub. When not defined will fallback to GITHUB_TOKEN |
```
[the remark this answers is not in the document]
luislard: updated

.github/workflows/automatic-release.yml
```diff
@@ -57,5 +63,5 @@ jobs:
 
       - name: Release
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.USER_TOKEN != '' && secrets.USER_TOKEN || secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
luislard: not sure about it, to be more accurate should be: READ_WRITE_TOKEN, and what about the different areas mentioned here? https://github.com/semantic-release/github#github-authentication 

Maybe **GH_AUTH_TOKEN** ?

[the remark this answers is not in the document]
luislard: Regarding the proposed logic, I am not adjusting to the proposed because the current one is leveraging the weird ternary operator from GH actions: see: https://docs.github.com/en/actions/learn-github-actions/expressions#example 

Current logic explanation:
If USER_TOKEN is not empty then return me that value, else the GITHUB_TOKEN

Proposed logic explanation:
(secrets.USER_TOKEN && secrets.USER_TOKEN != '') -> since the second part is an expression and not a value I think GH will resolve to a boolean value if the result is false will return GITHUB_TOKEN. The other scenario is -> if USER_TOKEN is truthy then it will return which means the expression `secrets.USER_TOKEN != ''` is not considered. 

Am I missing something?

.github/workflows/automatic-release.yml
```diff
@@ -7,6 +7,10 @@ on:
         default: 18
         required: false
         type: string
+    secrets:
+      USER_TOKEN:
+        description: Authentication token with write permission needed by the release bot.
```
tyrann0us, line 12:
```suggestion
        description: Authentication token with write permission needed by the release bot (falls back to GITHUB_TOKEN).
```

docs/automatic-release.md
```diff
@@ -40,9 +40,10 @@ jobs:
 
 #### Inputs
 
-| Name           | Default | Description                                          |
-|----------------|---------|------------------------------------------------------|
-| `NODE_VERSION` | 18      | Node version with which the release will be executed |
+| Name           | Required | Default | Description                                                                                |
+|----------------|----------|------------------|--------------------------------------------------------------------------------------------|
+| `NODE_VERSION` | false    |18               | Node version with which the release will be executed                                       |
+| `USER_TOKEN`   | false    |''          | Authentication token with write permission needed by the release bot. When not defined will fallback to GITHUB_TOKEN |
```
tyrann0us, line 46:
```suggestion
| `ACCESS_TOKEN`   | false    |''          | Authentication token with write permission needed by the release bot (falls back to GITHUB_TOKEN) |
```
Not sure about `ACCESS_TOKEN` or if we just should also use `GITHUB_USER_TOKEN` in the documentation.

Also, please update the Markdown table structure. Thanks!

docs/automatic-release.md
```diff
@@ -40,9 +40,10 @@ jobs:
 
 #### Inputs
 
-| Name           | Default | Description                                          |
-|----------------|---------|------------------------------------------------------|
-| `NODE_VERSION` | 18      | Node version with which the release will be executed |
+| Name                | Required | Default | Description                                                                                       |
+|---------------------|----------|---------|---------------------------------------------------------------------------------------------------|
+| `NODE_VERSION`      | false    |18       | Node version with which the release will be executed                                              |
+| `GITHUB_USER_TOKEN` | false    |''       | Authentication token with write permission needed by the release bot (falls back to GITHUB_TOKEN) |
```
tyrann0us, line 46:
```suggestion
| `GITHUB_USER_TOKEN` | false    |`''`     | Authentication token with write permission needed by the release bot (falls back to GITHUB_TOKEN) |
```

tyrann0us (approved):
Left one last minor suggestion. Apart from that LGTM. Thanks for working on this! 🙏🏽

Chrico (approved):
LGTM. 👍🏻

PR #90, opened 2023-11-22
fix: fixes Compile assets not respecting package manager when using npm

luislard:

Fixes wrong package manager used when compile assets step happens

When selecting **npm** as package manager the compile assets step uses **yarn**

When selecting **npm** as package manager the compile assets step uses **npm**

No

**Other information**:

### DIFF
--- a/.github/workflows/build-and-push-assets.yml
+++ b/.github/workflows/build-and-push-assets.yml
@@ -152,7 +152,7 @@ jobs:
         run: ${{ format('{0} {1}', inputs.PACKAGE_MANAGER, env.ARGS) }}
 
       - name: Compile assets
-        run: ${{ ((env.PACKAGE_MANAGER == 'yarn') && 'yarn') || 'npm run' }} ${{ env.COMPILE_SCRIPT }}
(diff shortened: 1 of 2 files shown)


tyrann0us (commented):
According to a quick [search](https://github.com/search?q=repo%3Ainpsyde%2Freusable-workflows+%22%27yarn%27%29+%26%26+%27yarn%27%29%22&type=code), the wrong brackets are used in multiple workflow files, of which `tests-unit-js.yml` is the only non-deprecated workflow. Therefore, please apply the change there, too: https://github.com/inpsyde/reusable-workflows/blob/70ffd21ec18900450ee46032b4966048b0cdeec6/.github/workflows/tests-unit-js.yml#L92

Also, according to the pathfinder rules, please remove the `PACKAGE_MANAGER` env variable here because : https://github.com/inpsyde/reusable-workflows/blob/dcf1691bff2af6c564cf83f781f7896a87723d48/.github/workflows/build-and-push-assets.yml#L81
The input defaults to `yarn` already; there is no need to define the fallback twice. Also, the env variable is only used in the changed line (which needs to change from `env.PACKAGE_MANAGER` to `inputs.PACKAGE_MANAGER`, of course).

Thanks!

tyrann0us (approved):
Haven't tested this, but code-wise LGTM. Thanks for fixing this! 🙏🏽

PR #109, opened 2024-02-09
docs(build-and-push-assets): `cancel-in-progress` usage is discouraged

shvlv:

Documentation.

We suggest using `concurrency.cancel-in-progress` in the consuming workflow for the `build-and-push` assets approach.

The problem here is the workflow run push at the same moment. It causes the next workflow run. But the original workflow may have to do some job after pushing (for instance, remove a temporary branch after tag moving). So, the safest way is to allow the original job to be completed and run the second job after.

No.

**Other information**:

### DIFF
--- a/docs/build-and-push-assets.md
+++ b/docs/build-and-push-assets.md
@@ -76,7 +76,8 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


tyrann0us (commented):
Does it make sense to remove the entire `concurrency` property from the documentation? `cancel-in-progress: false` should be the default, so setting up concurrency only to use a default value might confuse users.

tyrann0us (approved):
Thanks!

PR #111, opened 2024-02-26
Change default `@wordpress/scripts` flag values

### LINKED ISSUE #110: [Feature Request]: Evaluate possible default flag values for `@wordpress/scripts`

The `wp-scripts-lint.yml` workflow does not hardcode default flag values, so whenever a `*LINT_ARGS` input is used, calling workflows must pass the default value, too. This is verbose, resulting in a lot of copying & pasting. Also, it's very error-prone because users just don't know they must add default values after their custom flags.

It would be better if the reusable workflow would hardcode sensible default flag values. These would be appended after the `*LINT_ARGS` input.

Judging from this search: https://github.com/search?q=org%3Ainpsyde+OR+org%3AInpsyde-Global-Service-Provider+wp-scripts-lint+ARGS+language%3AYAML&type=code, the most often repeated flags are the following:
* `MARKDOWNLINT_ARGS`: `--ignore node_modules`
* `STYLELINT_ARGS`: `--formatter github`

Alternatively, we would update the documentation to warn users to always add defaults.

### Additional context

It is unclear what value is used when a flag is both set with an input and with the following hardcoded flag.

### Code of Conduct

Edit:
`MARKDOWNLINT_ARGS`
* Removed `--ignore LICENSE.md` because "modern" packages ship an extension-less `LICENSE` file
* `--ignore node_modules` might be eventually fixed upstream; see https://github.com/igorshubovych/markdownlint-cli/pull/460

tyrann0us:

Feature

There are two separate issues with flags passed to `@wordpress/scripts` commands:
1. **Stylelint**
  The `STYLELINT_ARGS` input defaults to `--formatter github`, so whenever a calling workflow uses this input, it must also provide this flag.
2. **Markdownlint**
  The `MARKDOWNLINT_ARGS` input defaults to `--ignore LICENSE.md --ignore node_modules`. The latter was only introduced (https://github.com/inpsyde/reusable-workflows/commit/c5632852cd13300016f3a36154ead8d7537bf498) because `@wordpress/scripts`' default behavior of ignoring the `node_modules/` folder is overridden when an `--ignore` flag is used. However, adding the first flag was actually an error because (current) packages don't use `LICENSE.md` files but an extension-free plain text `LICENSE` file.

This PR introduces two changes: Both the `STYLELINT_ARGS` and the `MARKDOWNLINT_ARGS` inputs default to an empty string.

For Stylelint, the `--formatter github` flag is _hardcoded_ appended to the command. This means calling workflows using `STYLELINT_ARGS` no longer have to pass this flag.

For Markdownlint, `@wordpress/scripts` will now ignore `node_modules/` by default if no `--ignore` flag is passed via the `MARKDOWNLINT_ARGS`.

No.

**Other information**:
Fixes #110

### DIFF
--- a/docs/wp-scripts.md
+++ b/docs/wp-scripts.md
@@ -33,7 +33,9 @@ jobs:
 | `MARKDOWNLINT_ARGS`     | `''`                           | Set of arguments passed to `wp-script lint-md-docs`                               |
 | `PACKAGE_JSONLINT_ARGS` | `''`                           | Set of arguments passed to `wp-scripts lint-pkg-json`                             |
 
-> :information_source: **By default, "pkg-json" is not part of the `LINT_TOOLS` input.**
+> :information_source: **By default, "pkg-json" is not part of the `LINT_TOOLS` input.**  
+> :information_source: **The `--formatter github` flag is hardcoded into the `wp-script lint-style`
+command; it must not be passed via `STYLELINT_ARGS`.**
 
 #### Secrets
 
@@ -67,6 +69,7 @@ jobs:
 ---
 **Note**
 
-Stylelint [requires quotes](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-scripts/#lint-style) around file glob patterns.
+Stylelint [requires quotes](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-scripts/#lint-style)
+around file glob patterns.
 
 ---


Chrico (approved):
LGTM 💪🏻

shvlv (approved):
Thanks!

.github/workflows/wp-scripts-lint.yml
```diff
@@ -135,7 +132,7 @@ jobs:
 
       - name: Lint style files
         if: ${{ contains(fromJSON(inputs.LINT_TOOLS), 'style') }}
-        run: ./node_modules/.bin/wp-scripts lint-style ${{ inputs.STYLELINT_ARGS }}
+        run: ./node_modules/.bin/wp-scripts lint-style ${{ inputs.STYLELINT_ARGS }} --formatter github
```
shvlv, line 135: In theory, this should work.

```suggestion
        run: echo ${INPUT_STYLELINT_ARGS//'--formatter github'/} | xargs ./node_modules/.bin/wp-scripts lint-style --formatter github
```
 The single problem could be https://github.com/orgs/community/discussions/27088. So the direct destruction might be required with:
```yaml
env:
   INPUT_STYLELINT_ARGS: ${{ inputs.STYLELINT_ARGS }}
```

tyrann0us: We discussed this internally and agreed not to workaround cases where `--formatter` is passed in `INPUT_STYLELINT_ARGS` because only very few packages are affected: https://github.com/search?q=org%3Ainpsyde+OR+org%3AInpsyde-Global-Service-Provider+STYLELINT_ARGS+--formatter&type=code.
These will be updated once they are worked on next.

Chrico (approved):
As discussed today (27.02.2024), we are not going to check for `--formatter` and hardcode it for stylelint to `github`. In case, in future, someone wants to choose between formatters, we are going to update that action and check via `contains()` the input and only append when a formatter is not set.

PR #115, opened 2024-03-12
Add solution about pre-releases on non-standard branches

dottxado:

Feature

Add the possibility to `-built` branches to be included in the automatic release process without generating invalid tag names for Composer

No

**Other information**:

### DIFF
--- a/docs/automatic-release.md
+++ b/docs/automatic-release.md
@@ -40,6 +40,48 @@ jobs:
     uses: inpsyde/reusable-workflows/.github/workflows/automatic-release.yml@main
 ```
 
+## Pre-releases on non-standard branches
+
+Composer has some boundaries on tag names, because it uses under the hood the PHP function `version_compare()`.
+Semantic-release doesn't account for these boundaries, so it can happen a created tag is not valid for Composer packages.
+A solution would be to add the [release.config.js file](https://github.com/inpsyde/reusable-workflows/blob/main/templates/automatic-release/release.config.js) in your repository and modify the `tagName` format accordingly.
+
+For example, using the [build-and-push-assets workflow](https://github.com/inpsyde/reusable-workflows/blob/main/.github/workflows/build-and-push-assets.yml) with a setting like this
+```yml
+name: Build and push assets
+
+on:
+  workflow_dispatch:
+  push:
+    branches: ['main', 'beta', 'alpha']
+
+jobs:
+  build-assets:
+    uses: inpsyde/reusable-workflows/.github/workflows/build-and-push-assets.yml@main
+    with:
+      BUILT_BRANCH_SUFFIX: "-built"
+```
+the `-built` branch is created after the building step of the assets, and you can run the automatic release on the same branch
+
+```yml
+name: Release
+on:
+  push:
+    branches:
+      - 'main-built'
+      - 'beta-built'
+      - 'alpha-built'
+jobs:
+  release:
+    uses: inpsyde/reusable-workflows/.github/workflows/automatic-release.yml@main
+```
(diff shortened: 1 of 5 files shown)


tyrann0us (commented):
According to [this search](https://github.com/search?q=BUILT_BRANCH_SUFFIX+language%3AYAML&type=code&l=YAML), the value of `BUILT_BRANCH_SUFFIX` is always `-built`. So, while it is configurable in theory, we don't use other values. (It might make sense to make this a boolean `USE_BUILT_BRANCH` with hardcoded value, but that's another topic.)

This makes me wonder whether we could put the `tagFormat` change in the `release.config.js` template, potentially saving users from maintaining this file in their repository if the only change is the `tagFormat` property:
```js
tagFormat: "<%- version.replace('-built', '') %>"
```

This would make adding this note to the documentation superfluous (while we should add an explanatory comment to the file). And for packages not using the built branch approach, no harm is done.
What do you think?
CC @shvlv

dottxado:
I partially agree.
If we include this setting in our template, we should also include the `-built` branches, otherwise it is totally useless.

So the template should be modified like this:

```
module.exports = {
  "branches": [
    "main",
    "main-built",
    "next",
    "next-built",
    {
      "name": "beta",
      "prerelease": true
    },
   {
      "name": "beta-built",
      "prerelease": true
    },
    {
      "name": "alpha",
      "prerelease": true
    },
   {
      "name": "alpha-built",
      "prerelease": true
    }
  ],
  "plugins": [
    "@semantic-release/commit-analyzer",
    "@semantic-release/release-notes-generator",
    "@semantic-release/changelog",
    ["@semantic-release/npm", {
      "tarballDir": "release",
      "npmPublish": false
    }],
    ["@semantic-release/exec", {
      "prepareCmd": "[ -f index.php ] || [ -f style.css ] && sed -i \"s/Version:.*/Version:     ${nextRelease.version}/\" index.php style.css || true",
    }],
    "@semantic-release/github",
    ["@semantic-release/git", {
      "assets": ["CHANGELOG.md", "package-lock.json", "package.json", "composer.json", "style.css", "index.php"],
      "message": "chore(release): \${nextRelease.version} [skip ci]\n\n\${nextRelease.notes}"
    }]
  ],
  "preset": "angular",
  "tagFormat": "<%- version.replace('-built', '') %>"
}
```

For my team that uses frequently these two workflows in combination, this change would be awesome, because, as you said, we have to maintain a local copy only to add the `-built` branches.

tyrann0us:
> If we include this setting in our template, we should also include the -built branches, otherwise it is totally useless.

Yes, good point. I'm okay with extending the configuration file like this. Please note that it should be possible to use wildcards for branch names according to the documentation (https://github.com/semantic-release/semantic-release/blob/master/docs/usage/workflow-configuration.md#name), so you could do something like `*-built` (untested!).

@Chrico, are you okay with hardcoding a value (`-built`) that is actually configurable (but nobody uses a different value; see above)?

dottxado:
I add another point about wildcard branches :smile: since Composer accepts only `alpha`, `beta`, `RC` and `pl` strings in the tag name, having a wildcard can spark the possibility of having a `foo-built` pre-release branch and resulting in a `1.0.0-foo.1` tag. Maybe if we want to enforce a working standard, having the branch listed can be a good way? :woman_shrugging:

docs/automatic-release.md
tyrann0us, line 1: I wonder whether we should add this explanation to the documentation. I'm not strictly against it, but it could potentially confuse users. Also, you mention that the file should be copied from this repository if needed, but actually, the CLI should be responsible for providing boilerplate code, IMO.
Can we just revert the change to this file? 🙈

tyrann0us (approved):
Thanks for incorporating all the changes. LGTM! 😉

PR #116, opened 2024-03-18
Add main file name customization on automatic release

dottxado:

Feature

The automatic release feature can be added only on plugins where the main plugin file is `index.php`.

The main file name is customizable.

No

**Other information**:
This feature opens the possibility of using this automation for plugins built with different standards on file names (like the WordPress standard).

### DIFF
--- a/.github/workflows/automatic-release.yml
+++ b/.github/workflows/automatic-release.yml
@@ -58,7 +58,7 @@ jobs:
         if: ${{ env.HAS_CONFIG == 'false' }}
         run: |
           cp ${GITHUB_WORKSPACE}/workflow-repo/templates/automatic-release/release.config.js .
-          FILE=$(find . -maxdepth 1 -type f -name '*.php' -exec grep -l 'Plugin Name:' {} + | xargs -I{} basename {}) || true
+          FILE=$(find . -maxdepth 1 -type f -name '*.php' -exec grep -l 'Plugin Name:' {} + | xargs -I{} basename {}) || false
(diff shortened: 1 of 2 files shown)


.github/workflows/automatic-release.yml
```diff
@@ -54,10 +60,11 @@ jobs:
           repository: inpsyde/reusable-workflows
           path: workflow-repo
 
-      - name: Add release.config.js file if not provided
+      - name: Add release.config.js file if not provided and customize it
         if: ${{ env.HAS_CONFIG == 'false' }}
         run: |
           cp ${GITHUB_WORKSPACE}/workflow-repo/templates/automatic-release/release.config.js .
+          sed -i 's/index\.php/${{inputs.PLUGIN_MAIN_FILENAME}}/g' release.config.js
```
tyrann0us, line 67: We agreed on something different in the last jour fixe, but I couldn't resist testing an automated approach to identify the main plugin file name, which would prevent us from introducing the `PLUGIN_MAIN_FILENAME` input.

The following bash command identifies the main plugin file by the (mandatory) `Plugin Name:` plugin header and replaces `index.php` with it:
```bash
FILE=$(find . -maxdepth 1 -type f -name '*.php' -exec grep -l 'Plugin Name:' {} + | xargs -I{} basename {})
sed "s/index\.php/$FILE/g" -i release.config.js
```

I tested it locally, and it works like a charm (thanks, ChatGPT! 😆). If you agree, this would be the only required change in this PR. 😉
CC @Chrico

tyrann0us (approved):

PR #108, opened 2024-02-06
Default to PHP 8.2

tyrann0us:
**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes/features)
- [x] Docs have been added/updated (for bug fixes/features)

**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Refactor

**What is the current behavior?** (You can also link to an open issue here)
All PHP-based reusable workflows default to PHP 8.0, which reached EOL on 2023-11-26.

**What is the new behavior (if this is a feature change)?**
Default all PHP-based reusable workflows to PHP 8.2, which will reach EOL on 2025-12-08.

**Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
Yes. All calling workflows not specifying a PHP version with the `PHP_VERSION` input will run on PHP 8.2, potentially causing the workflow to fail.

**Other information**:
Closes #44.

PR #117, opened 2024-03-19
Multi step plugin archive

Biont:

Refactoring

`build-plugin-archive.yml` uses a single php version for bundling dependencies and executing build tools.
WP's committment to supporting php down to 7.0 makes it increasingly cumbersome to support **"one"** *php platform requirement* for production AND development/building.

We have been accepting this limitation in order to streamline our packaging across most of our projects - and it's been a very solid base. But the fact remains that the php ecosystem is moving on while the php7 support requirement for WP packages is not going anywhere. Since this gap is only getting wider, we need to find a better approach

The logic of the workflow has been split into 3 parts:

1. **Production dependencies** - checkout the repo and `composer install` using the ancient PHP version
2. **Build steps** - execute all relevant build tools using a much more modern php version. We also no longer expect these to be present in dev dependencies: If we find a config, we install the tool globally. 
3. **Archive creation** - add final touches and create the resulting archive

Note: There is no strict technical requirement for step 3 to be in a separate job. It adds a little overhead, but I think at this point, it adds a nice separation of concerns in that it keeps a few environment variables and data handling out of the other steps.

The individual jobs each produce their own "interim" artifact which is then picked up by the next one. This makes it easy to spot and debug where something went wrong:

screenshot_2024-03-20-120520

---

This change frees us from having to use EOL versions of common tools and also paves the way for modernization of our own tools, for example [dropping PHP 7 in Composer Asset Compiler](https://github.com/inpsyde/composer-asset-compiler/pull/23). It also allows us to introduce third-party tools like *php-scoper* with ease since we are no longer subject to the php version constraints of individual projects.

No

**Other information**:

### DIFF
--- a/.github/workflows/build-plugin-archive.yml
+++ b/.github/workflows/build-plugin-archive.yml
@@ -208,7 +208,7 @@ jobs:
       - name: Install Composer Asset Compiler
         if: steps.composer-tools.outputs.assets-compiler == '0' && steps.check-for-package.outputs.has_package == 'true'
         run: |
-          composer global config --no-plugins allow-plugins.inpsyde/composer-assets-compiler
+          composer global config --no-plugins --no-interaction allow-plugins.inpsyde/composer-assets-compiler true
           composer global require inpsyde/composer-assets-compiler
       - name: Run Composer Asset Compiler
         if: steps.composer-tools.outputs.assets-compiler == '0' && steps.check-for-package.outputs.has_package == 'true'
@@ -217,7 +217,7 @@ jobs:
       - name: Install WordPress Translation Downloader
         if: steps.composer-tools.outputs.translation-downloader == '0'
         run: |
-          composer global config --no-plugins allow-plugins.inpsyde/wp-translation-downloader
+          composer global config --no-plugins --no-interaction allow-plugins.inpsyde/wp-translation-downloader true
           composer global require inpsyde/wp-translation-downloader
       - name: Run WordPress Translation Downloader
         if: steps.composer-tools.outputs.translation-downloader == '0'
--- a/.github/workflows/build-plugin-archive.yml
+++ b/.github/workflows/build-plugin-archive.yml
@@ -24,12 +24,17 @@ on:
         required: false
         type: string
       COMPOSER_ARGS:
-        description: Set of arguments passed to Composer.
+        description: Set of arguments passed to Composer when gathering production dependencies.
         default: '--no-dev --prefer-dist --optimize-autoloader'
         required: false
         type: string
       PHP_VERSION:
-        description: PHP version to use during packaging.
+        description: PHP version to use when gathering production dependencies.
+        default: "8.0"
+        required: false
+        type: string
+      PHP_VERSION_BUILD:
+        description: PHP version to use when executing build tools.
         default: "8.0"
         required: false
         type: string
@@ -251,7 +256,7 @@ jobs:
       - name: Set up PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '8.3'
+          php-version: ${{ inputs.PHP_VERSION_BUILD }}
           tools: wp-cli
 
       - name: Install dist-archive command
--- a/.github/workflows/build-plugin-archive.yml
+++ b/.github/workflows/build-plugin-archive.yml
@@ -200,15 +200,13 @@ jobs:
           composer global require inpsyde/composer-assets-compiler
           composer compile-assets ${{ inputs.COMPILE_ASSETS_ARGS }}
 
-      - name: Install WordPress Translation Downloader
+      - name: Install and run WordPress Translation Downloader
         if: steps.composer-tools.outputs.translation-downloader == '0'
         run: |
           composer global config --no-plugins --no-interaction allow-plugins.composer/installers true
           composer global config --no-plugins --no-interaction allow-plugins.inpsyde/wp-translation-downloader true
           composer global require inpsyde/wp-translation-downloader
-      - name: Run WordPress Translation Downloader
-        if: steps.composer-tools.outputs.translation-downloader == '0'
-        run: composer wp-translation-downloader:download
+          composer wp-translation-downloader:download
 
       - name: Set artifact name
         id: set-artifact-name
--- a/.github/workflows/build-plugin-archive.yml
+++ b/.github/workflows/build-plugin-archive.yml
@@ -139,6 +139,7 @@ jobs:
   run-build-tools:
     timeout-minutes: 5
     runs-on: ubuntu-latest
+    needs: checkout-dependencies
     env:
       NODE_OPTIONS: ${{ inputs.NODE_OPTIONS }}
       NODE_AUTH_TOKEN: ${{ secrets.NPM_REGISTRY_TOKEN }}
@@ -206,6 +207,7 @@ jobs:
   create-plugin-archive:
     timeout-minutes: 5
     runs-on: ubuntu-latest
+    needs: run-build-tools
     env:
       ARCHIVE_NAME: ${{ inputs.ARCHIVE_NAME }}
     steps:
--- a/.github/workflows/build-plugin-archive.yml
+++ b/.github/workflows/build-plugin-archive.yml
@@ -79,6 +79,7 @@ on:
 
 jobs:
   checkout-dependencies:
+    name: Install production dependencies
     timeout-minutes: 5
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-plugin-archive.yml
```diff
@@ -118,38 +104,82 @@ jobs:
...
+          hasTranslateConfig(){
+            jq '.extra | has("wp-translation-downloader")' < composer.json >/dev/null 2>&1
+            local EXIT=$?
+            echo "$EXIT"
+          }
           hasDep(){
             composer show -i -N -D --strict "${1}" >/dev/null 2>&1
             local EXIT=$?
             echo "$EXIT"
           }
```
tyrann0us, line 170: `hasDep()` is no longer used.
```suggestion
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -118,38 +104,82 @@ jobs:
...
+        run: |
+          if [ "${{ inputs.PACKAGE_MANAGER }}" == 'npm' ] && { [ -f "${GITHUB_WORKSPACE}/package-lock.json" ] || [ -f "${GITHUB_WORKSPACE}/npm-shrinkwrap.json" ]; }; then
+            echo "NODE_CACHE_MODE=npm" >> $GITHUB_ENV
+          elif [ "${{ inputs.PACKAGE_MANAGER }}" == 'yarn' ] && [ -f "${GITHUB_WORKSPACE}/yarn.lock" ]; then
+            echo "NODE_CACHE_MODE=yarn" >> $GITHUB_ENV
+          else
+            echo "No lock files found or unknown package manager"
+          fi
 
       - name: Check for package.json
```
tyrann0us, line 184: I know this is existing code, but why do we need this step? We already check for the presence of the Composer Asset Compiler configuration; there's no valid reason for it to be present in a package that does _not_ have a `package.json` file.

Also, the `check-for-package` step ID just confused me while reading the code. It is always a package; what we do here is to check if it is an _npm_ package or rather if it has a `package.json` file. If you decide to keep this step, can we at least improve the ID?

.github/workflows/build-plugin-archive.yml
```diff
@@ -168,32 +198,77 @@ jobs:
...
           cache: ${{ env.NODE_CACHE_MODE }}
 
+      - name: Install Composer Asset Compiler
+        if: steps.composer-tools.outputs.assets-compiler == '0' && steps.check-for-package.outputs.has_package == 'true'
+        run: |
+          composer global config --no-plugins --no-interaction allow-plugins.inpsyde/composer-assets-compiler true
+          composer global require inpsyde/composer-assets-compiler
       - name: Run Composer Asset Compiler
         if: steps.composer-tools.outputs.assets-compiler == '0' && steps.check-for-package.outputs.has_package == 'true'
         run: composer compile-assets ${{ inputs.COMPILE_ASSETS_ARGS }}
```
tyrann0us, line 208: Can we merge this into one step?
```suggestion
      - name: Install and run Composer Asset Compiler
        if: steps.composer-tools.outputs.assets-compiler == '0'
        run: |
          composer global config --no-plugins --no-interaction allow-plugins.inpsyde/composer-assets-compiler true
          composer global require inpsyde/composer-assets-compiler
          composer compile-assets ${{ inputs.COMPILE_ASSETS_ARGS }}
```

Note: I already removed the `has_package` check (see comment above).

.github/workflows/build-plugin-archive.yml
```diff
@@ -168,32 +198,77 @@ jobs:
...
 
+      - name: Install WordPress Translation Downloader
+        if: steps.composer-tools.outputs.translation-downloader == '0'
+        run: |
+          composer global config --no-plugins --no-interaction allow-plugins.composer/installers true
+          composer global config --no-plugins --no-interaction allow-plugins.inpsyde/wp-translation-downloader true
+          composer global require inpsyde/wp-translation-downloader
       - name: Run WordPress Translation Downloader
         if: steps.composer-tools.outputs.translation-downloader == '0'
         run: composer wp-translation-downloader:download
```
tyrann0us, line 218: Same as above.
```suggestion
      - name: Install and run WordPress Translation Downloader
        if: steps.composer-tools.outputs.translation-downloader == '0'
        run: |
          composer global config --no-plugins --no-interaction allow-plugins.composer/installers true
          composer global config --no-plugins --no-interaction allow-plugins.inpsyde/wp-translation-downloader true
          composer global require inpsyde/wp-translation-downloader
          composer wp-translation-downloader:download
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -25,7 +25,7 @@ on:
         type: string
       COMPOSER_ARGS:
         description: Set of arguments passed to Composer.
-        default: '--no-dev --no-scripts --prefer-dist --optimize-autoloader'
+        default: '--no-dev --prefer-dist --optimize-autoloader'
         required: false
         type: string
       PHP_VERSION:
```
tyrann0us, line 31: The description needs to be updated, I guess.

.github/workflows/build-plugin-archive.yml
```diff
@@ -168,32 +198,77 @@ jobs:
...
       - name: Add commit hash to plugin header
         run: 'sed -Ei "s/SHA: .*/SHA: ${GIT_SHA}/g" ${{ inputs.PLUGIN_MAIN_FILE }}'
 
       - name: Set plugin version header
         run: 'sed -Ei "s/Version: .*/Version: ${PLUGIN_VERSION}/g" ${{ inputs.PLUGIN_MAIN_FILE }}'
 
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.3'
```
tyrann0us, line 263: Let's please not hardcode a PHP version here. I understand you want to use a different version than `${{ inputs.PHP_VERSION }}`, but in this case, please create an additional input (which ideally would default to PHP 8.0 to be aligned with all other PHP version inputs; see #108).

tyrann0us (changes requested):

.github/workflows/build-plugin-archive.yml
Biont: Yes but I am letting go of this awesome piece of scripting with a heavy heart, just so you know

.github/workflows/build-plugin-archive.yml
Biont: Okay makes sense if you prefer it that way

.github/workflows/build-plugin-archive.yml
Biont: Ha! I was actually just trying to see if the `tools:` section would make life easier by setting up `php-scoper` in the `PATH` as opposed to my previous hamfisted approach of finding the executable.
I even recall thinking "maybe this should be a new input" but then forgot. I'll add this

.github/workflows/build-plugin-archive.yml
Biont: Done

.github/workflows/build-plugin-archive.yml
Biont: Done

.github/workflows/build-plugin-archive.yml
Biont: I removed it. Thanks for pointing this out

docs/archive-creation.md
```diff
@@ -46,7 +46,8 @@ jobs:
 | `NPM_REGISTRY_DOMAIN` | `"https://npm.pkg.github.com/"`                               | Domain of the private npm registry                                                             |
 | `PACKAGE_MANAGER`     | `"yarn"`                                                      | Package manager with which the dependencies should be installed (`npm` or `yarn`)              |
 | `COMPOSER_ARGS`       | `'--no-dev --no-scripts --prefer-dist --optimize-autoloader'` | Set of arguments passed to Composer                                                            |
```
tyrann0us, line 48:
```suggestion
| `COMPOSER_ARGS`       | `'--no-dev --no-scripts --prefer-dist --optimize-autoloader'` | Set of arguments passed to Composer when gathering production dependencies |
```
Please ensure to align the table.

tyrann0us (approved):
Thanks for incorporating the changes so quickly. LGTM. Left a minor comment that doesn't stop me from approving. 💪🏽

docs/archive-creation.md
Biont: I did reformat using the PHPStorm inspection, but yeah I forgot to update the description for `COMPOSER_ARGS`. Thanks for that!

PR #129, opened 2024-06-03
Update build-plugin-archive.yml to Include Interim Steps and Clean Up

mmaymo:
This PR modifies the `build-plugin-archive.yml` workflow by introducing interim steps for modifying the plugin header and cleaning up after the WP-CLI command. The changes ensure that the necessary modifications are made within an interim directory and the directory is cleaned up properly.

#### Changes Made:
1. **Add Commit Hash to Plugin Header:**
   - Changed the command to run in an interim directory (`interim-deps`).
   - Updated the script to move to the interim directory before executing the `sed` command.
   
   ```yaml
   - name: Add commit hash to plugin header
     run: |
       cd interim-deps
       sed -Ei "s/SHA: .*/SHA: ${GIT_SHA}/g" ${{ inputs.PLUGIN_MAIN_FILE }}
   ```

2. **Set Plugin Version Header:**
   - Similar to the commit hash step, this also runs in the interim directory.
   - Updated the script to move to the interim directory before executing the `sed` command.
   
   ```yaml
   - name: Set plugin version header
     run: |
       cd interim-deps
       sed -Ei "s/Version: .*/Version: ${{ inputs.PLUGIN_VERSION }}/g" ${{ inputs.PLUGIN_MAIN_FILE }}
   ```

3. **Run WP-CLI Command:**
   - Added an interim directory cleanup step before executing the WP-CLI command.
   
   ```yaml
   - name: Run WP-CLI command
     run: |
       rm -rf interim-deps/.git
       wp dist-archive ./archive.zip --plugin-dirname=${{ steps.plugin-folder-name.outputs.plugin-folder-name }}
   ```

4. **Unzip Archive to Dist:**
   - No changes in this step but it follows the updated steps to ensure clean-up.
   
   ```yaml
   - name: Unzip archive to dist
   ```

#### Testing

- Verified that the workflow runs successfully and the `sed` commands update the plugin header as expected.
- Ensured that the WP-CLI command produces the desired archive without any leftover files from the interim directory.

Please review the changes and let me know if there are any questions or further adjustments needed.

### DIFF
--- a/.github/workflows/build-plugin-archive.yml
+++ b/.github/workflows/build-plugin-archive.yml
@@ -208,7 +208,7 @@ jobs:
       - name: Upload artifact
         uses: actions/upload-artifact@v4
         with:
-          name: ${{ steps.run-build-tools.outputs.artifact }}
+          name: ${{ steps.set-artifact-name.outputs.artifact }}
           path: |
             .
             !**/node_modules
@@ -266,7 +266,7 @@ jobs:
           ${{ inputs.PRE_SCRIPT }}
 
       - name: Run WP-CLI command
-        run: wp dist-archive ./interim-built ./archive.zip --plugin-dirname=${{ steps.plugin-folder-name.outputs.plugin-folder-name }}
+        run: wp dist-archive ./interim-deps ./archive.zip --plugin-dirname=${{ steps.plugin-folder-name.outputs.plugin-folder-name }}
 
         # GitHub Action artifacts would otherwise produce a zip within a zip
       - name: Unzip archive to dist/
--- a/.github/workflows/build-plugin-archive.yml
+++ b/.github/workflows/build-plugin-archive.yml
@@ -263,7 +263,6 @@ jobs:
 
       - name: Run WP-CLI command
         run: |
-          ls -laR
           rm -rf interim-deps/.git
           wp dist-archive ./interim-deps ./archive.zip --plugin-dirname=${{ steps.plugin-folder-name.outputs.plugin-folder-name }}
 
--- a/.github/workflows/build-plugin-archive.yml
+++ b/.github/workflows/build-plugin-archive.yml
@@ -242,10 +242,14 @@ jobs:
         run: echo "archive-name=${ARCHIVE_NAME:-${{ github.event.repository.name }}}" >> $GITHUB_OUTPUT
 
       - name: List files in current directory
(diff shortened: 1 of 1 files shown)


.github/workflows/build-plugin-archive.yml
```diff
@@ -239,10 +239,14 @@ jobs:
         run: echo "archive-name=${ARCHIVE_NAME:-${{ github.event.repository.name }}}" >> $GITHUB_OUTPUT
 
       - name: Add commit hash to plugin header
-        run: 'sed -Ei "s/SHA: .*/SHA: ${GIT_SHA}/g" ${{ inputs.PLUGIN_MAIN_FILE }}'
+        run: |
+          cd interim-deps
```
Biont, line 243: I think using `working_dir` would be a little cleaner.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsworking-directory

.github/workflows/build-plugin-archive.yml
```diff
@@ -239,10 +239,14 @@ jobs:
...
+          sed -Ei "s/SHA: .*/SHA: ${GIT_SHA}/g" ${{ inputs.PLUGIN_MAIN_FILE }}
 
       - name: Set plugin version header
-        run: 'sed -Ei "s/Version: .*/Version: ${PLUGIN_VERSION}/g" ${{ inputs.PLUGIN_MAIN_FILE }}'
+        run: |
+          cd interim-deps
```
Biont, line 248: Same as above

.github/workflows/build-plugin-archive.yml
```diff
@@ -258,7 +262,9 @@ jobs:
           ${{ inputs.PRE_SCRIPT }}
 
       - name: Run WP-CLI command
-        run: wp dist-archive . ./archive.zip --plugin-dirname=${{ steps.plugin-folder-name.outputs.plugin-folder-name }}
+        run: |
+          rm -rf interim-deps/.git
```
Biont, line 266: I realize and respect that there likely isn't ever going to be a use-case for shipping the git clone in a build. 

I still wonder whether this should better be handled entirely by the `.distignore` file.

Biont (changes requested):

Biont (approved):
Yay

PR #134, opened 2024-06-20
Additional tools for plugin archives: php-scoper, rector

Biont:

feature

The existing multi-step plugin archive workflow can currently identify and execute the following tools
 - asset compilation via `inpsyde/composer-asset-compiler`
 - translation download via `inpsyde/wp-translation-downloader`

In addition, the following "post-processing" tools are introduced
 - namespace prefixing via `humbug/php-scoper`
 - transpilation of php8 code via `rectorphp/rector`

No

**Other information**:


.github/workflows/build-plugin-archive.yml
```diff
@@ -118,82 +109,174 @@ jobs:
...
 
-      - name: Install Composer dependencies without dev dependencies
-        # ⬑ Now we want to get rid of dev dependencies and most probably skip composer scripts
-        uses: ramsey/composer-install@v3
+      - name: Run PHP Scoper
+        if: steps.composer-tools.outputs.php-scoper == '0'
+        run: |
+          php-scoper add-prefix --force --output-dir build
+          composer --working-dir=build dump-autoload -o
+          sed -i "s/'__composer_autoload_files'/\'__composer_autoload_files_${{ github.sha }}'/g" "build/vendor/composer/autoload_real.php"
```
Biont, line 223: This one warrants some attention. When autoloading files, composer checks a hash of `<PACKAGE>+<PATH>` against a list in `$GLOBALs['__composer_autoload_files']`. If the hash is already present, the file will not be included.

Since composer does not care about the **file contents** in this hash, it will prevent the prefixed version of the file from being included (despite clearly being a different piece of code now). This essentially breaks "Inpsyde Assets" because it relies on autoloaded files, and -when prefixed- fails to instantiate if another instance is present on the WP installation.

Since the intent is clearly to implement "full prefixing" of the entire plugin (as opposed to scoping away only a few known incompatibilities), we do not expect major issues from potentially allowing duplicate includes through this workaround. Yet we obviously cannot rule them out either. I checked with colleagues and we're willing to dogfood this a little and accept that risk.

.github/workflows/build-plugin-archive.yml
```diff
@@ -202,16 +215,27 @@ jobs:
...
 
+      - name: Run PHP Scoper
+        if: steps.composer-tools.outputs.php-scoper == '0'
+        run: |
+          php-scoper add-prefix --force --output-dir build
+          composer --working-dir=build dump-autoload -o
+          sed -i "s/'__composer_autoload_files'/\'__composer_autoload_files_${{ github.sha }}'/g" "build/vendor/composer/autoload_real.php"
+      - name: Move unchanged code to build directory
+        if: steps.composer-tools.outputs.php-scoper != '0'
+        run: mv * build 2>/dev/null
```
AlexP11223, line 226: Looks like it does not work, probably because the `build` dir does not exist.

Fixed in another branch https://github.com/inpsyde/reusable-workflows/commit/3ba25630c7e7edc030f74a625618d37e4bb6280b
Not sure if there is a better way and if `shopt -s extglob` can cause issues.

.github/workflows/build-plugin-archive.yml
```diff
@@ -35,7 +35,7 @@ on:
         type: string
       PHP_VERSION_BUILD:
         description: PHP version to use when executing build tools.
-        default: "8.0"
+        default: "8.2"
```
tyrann0us, line 38: Can we leave the PHP update to the PR that touches all files, #108?
```suggestion
        default: "8.0"
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -147,6 +149,11 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: ${{ needs.checkout-dependencies.outputs.artifact }}
+      - name: Setup PHP with tools
```
tyrann0us, line 152: In all other workflows, we name this step "Set up PHP", regardless of whether or not we use it to install additional tools. Also, there's a new line above each step.
```suggestion

      - name: Set up PHP
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -147,6 +149,11 @@ jobs:
...
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ inputs.PHP_VERSION_BUILD }}
+          tools: humbug/php-scoper:dev-main
```
tyrann0us, line 156: Is there a particular reason for using `dev-main`? Unreleased, buggy changes on that branch could lead to hard-to-detect bugs in the generated code.
```suggestion
          tools: humbug/php-scoper

```
Also added a new line above the next step.

.github/workflows/build-plugin-archive.yml
```diff
@@ -202,16 +215,27 @@ jobs:
           composer global require inpsyde/wp-translation-downloader
           composer wp-translation-downloader:download
 
+      - name: Run PHP Scoper
```
tyrann0us, line 218: The official name is "PHP-Scoper"; see https://github.com/humbug/php-scoper.
```suggestion
      - name: Run PHP-Scoper
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -202,16 +215,27 @@ jobs:
...
+        if: steps.composer-tools.outputs.php-scoper == '0'
+        run: |
+          php-scoper add-prefix --force --output-dir build
```
tyrann0us, line 221: Can we align the flag syntax?
```suggestion
          php-scoper add-prefix --force --output-dir=build
```

.github/workflows/build-plugin-archive.yml
tyrann0us: While I'm okay with including this step, it is fragile. More importantly, I wonder whether this should not be a feature of PHP-Scoper instead. Does it make sense to open an issue there and try to get this "fixed" upstream?

.github/workflows/build-plugin-archive.yml
```diff
@@ -202,16 +215,27 @@ jobs:
...
+          composer --working-dir=build dump-autoload -o
+          sed -i "s/'__composer_autoload_files'/\'__composer_autoload_files_${{ github.sha }}'/g" "build/vendor/composer/autoload_real.php"
+      - name: Move unchanged code to build directory
```
tyrann0us, line 224: On the first read, the "unchanged" part of the step name is confusing, IMO. Would it be possible to run the previous step in the current directory instead of in `build/`? This would make this step superfluous. If not, let's try to improve the name.
```suggestion
      - name: Move code to the build directory

```
New line also added, again.

tyrann0us (changes requested):

.github/workflows/build-plugin-archive.yml
```diff
@@ -147,6 +149,11 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: ${{ needs.checkout-dependencies.outputs.artifact }}
+      - name: Setup PHP with tools
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ inputs.PHP_VERSION_BUILD }}
+          tools: humbug/php-scoper:dev-main, rector
```
tyrann0us, line 156: The previous conversation (https://github.com/inpsyde/reusable-workflows/pull/134#discussion_r1741806490) was hidden by merging #135; commenting this again so it doesn't get lost:
Is there a particular reason for using `dev-main`? Unreleased, buggy changes on that branch could lead to hard-to-detect bugs in the generated code.
```suggestion
          tools: humbug/php-scoper, rector

```
Also added a new line above the next step.

.github/workflows/build-plugin-archive.yml
Biont: https://github.com/humbug/php-scoper/issues/298
https://github.com/humbug/php-scoper/issues/1036

[There also used to be documentation about this](https://github.com/humbug/php-scoper/pull/1046/files), but it has been removed a while ago. Maybe this is worth another look..

.github/workflows/build-plugin-archive.yml
Biont: [There must have been a reason](https://github.com/inpsyde/reusable-workflows/pull/134/commits/946f4e425080eb9ab598334e8f0d2867f55d3960), but that was before the latest public release, so removing it should be fine - or at least worth trying

.github/workflows/build-plugin-archive.yml
Biont: What's happening here:

 - If php-scoper is in use, it has processed the code from `cwd` into `build/`
 - If php-scoper it **NOT** in use, then the code is _unchanged_ and `build/` is empty
 
 Therefore, we need this step to make sure that in all scenarios, we can continue with the workflow from `build/`

tyrann0us: Yes, yes, I understand the code. My question was if PHP-Scoper could work in the current directory, making this step superfluous for products not using the tool.

AlexP11223:
Added `include-hidden-files: true` in uploads, to fix this breaking change: https://github.com/actions/upload-artifact/issues/602

.github/workflows/build-plugin-archive.yml
Biont: I checked this again: This isn't actually an issue with php-scoper - it prefixes the files just fine. The problem is that *composer* refuses to load "the same file twice" and it decides that based on the relative file path, ignoring the actual file contents (which are different and non-conflicting now after scoping).

IIRC we tried to implement post-processing to salt the hashes via scope-config, but weren't able to come up with a universal and easy solution. So doing it in the GHA appeared like the logical next alternative.

@tyrann0us

.github/workflows/build-plugin-archive.yml
Biont: To my knowledge, php-scoper cannot work in-place, so the output dir must differ. An alternative could be to let scoper write to a temp directory and then replace the current dir's contents.
This might make it a bit clearer to chain additional tools since we do not switch directories at some point. But on a technical level, it's not much of a difference. What do you prefer?

tyrann0us: >What do you prefer?

Let's keep things as-is, but maybe change the step name as suggested.

.github/workflows/build-plugin-archive.yml
```diff
@@ -202,16 +224,37 @@ jobs:
...
+          rector
+          
+      - name: Run PHP-Scoper
+        if: steps.composer-tools.outputs.php-scoper == '0'
+        run: |
+          php-scoper add-prefix --force --output-dir=build
+          composer --working-dir=build dump-autoload -o
+          sed -i "s/'__composer_autoload_files'/\'__composer_autoload_files_${{ github.sha }}'/g" "build/vendor/composer/autoload_real.php"
+      - name: Move code to the `build/` directory
+
```
tyrann0us, line 239: Sorry, my suggestion placed the new line wrongly.
```suggestion

      - name: Move code to the `build/` directory
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -131,9 +133,10 @@ jobs:
             !./.git
             !./.ddev
             !./.github
+          include-hidden-files: true
   run-build-tools:
```
tyrann0us, line 137:
```suggestion

  run-build-tools:
```

.github/workflows/build-plugin-archive.yml
```diff
@@ -202,16 +224,37 @@ jobs:
...
         uses: actions/upload-artifact@v4
         with:
           name: ${{ steps.set-artifact-name.outputs.artifact }}
           path: |
-            .
-            !**/node_modules
+            build/
+            !build/**/node_modules
+          include-hidden-files: true
   create-plugin-archive:
```
tyrann0us, line 258:
```suggestion

  create-plugin-archive:
```

tyrann0us (approved):
Thanks for incorporating all the changes. Left a minor round of suggestions to get the new lines straight, but this doesn't stop me from approving.

tyrann0us (approved):

PR #135, opened 2024-07-03
Add Rector in archive creation

AlexP11223:

feature

Now the package creation workflow also runs [Rector](https://github.com/rectorphp/rector) if the config is present. This was needed for researching PHP 8 transpiling into PHP 7.4, but also can be useful for other tasks.

Increased the timeout to 10 minutes because sometimes it takes a bit more than 5 minutes. (Rector takes around 40 seconds in the WL project)

### DIFF
--- a/.github/workflows/build-plugin-archive.yml
+++ b/.github/workflows/build-plugin-archive.yml
@@ -164,12 +164,12 @@ jobs:
               echo "$EXIT"
               exit 0
             fi
-            jq '.extra | has("composer-asset-compiler")' < composer.json >/dev/null 2>&1
+            jq '.extra | has("composer-asset-compiler")' --exit-status < composer.json >/dev/null 2>&1
             local EXIT=$?
             echo "$EXIT"
           }
           hasTranslateConfig(){
-            jq '.extra | has("wp-translation-downloader")' < composer.json >/dev/null 2>&1
+            jq '.extra | has("wp-translation-downloader")' --exit-status < composer.json >/dev/null 2>&1
             local EXIT=$?
             echo "$EXIT"
           }
--- a/.github/workflows/build-plugin-archive.yml
+++ b/.github/workflows/build-plugin-archive.yml
@@ -153,7 +153,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ inputs.PHP_VERSION_BUILD }}
-          tools: humbug/php-scoper:dev-main
+          tools: humbug/php-scoper:dev-main, rector
       - name: Check optional Composer build tools
         id: composer-tools
         run: |
@@ -224,7 +224,7 @@ jobs:
       - name: Run Rector
         if: steps.composer-tools.outputs.rector == '0'
         run: |
-          vendor/bin/rector
+          rector
           
       - name: Run PHP Scoper
         if: steps.composer-tools.outputs.php-scoper == '0'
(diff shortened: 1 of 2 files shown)


Chrico (commented):
What is the use case of Rector? Isn't that something you only execute once and when the project is upgraded you can remove it? 👀 Right now, it seems that code is changed on the fly, but never tested before creating a release. So you build a release on something untested?

AlexP11223:
There are other PRs with workflows for testing the archive. :) (+ hopefully Playwright in the future)

For transpiling the idea is that we use the latest PHP for development, and then transform into 7.4 for release.
https://inpsyde.atlassian.net/browse/PROD-171?focusedCommentId=298937

Biont:
>So you build a release on something untested?

This here is a build/packaging workflow, not a release workflow. The produced builds are a prerequisite for further (automated&manual) testing by QA.

.github/workflows/build-plugin-archive.yml
```diff
@@ -223,7 +234,11 @@ jobs:
           sed -i "s/'__composer_autoload_files'/\'__composer_autoload_files_${{ github.sha }}'/g" "build/vendor/composer/autoload_real.php"
       - name: Move unchanged code to build directory
         if: steps.composer-tools.outputs.php-scoper != '0'
-        run: mv * build 2>/dev/null
+        run: |
+          shopt -s extglob
+          shopt -s dotglob
```
tyrann0us, line 239: Let's simplify this.
```suggestion
          shopt -s extglob dotglob
```

docs/archive-creation.md
```diff
@@ -7,10 +7,11 @@ To achieve that, the reusable workflow:
...
 3. Executes `inpsyde/wp-translation-downloader` if required & configured by the package
-4. Re-installs dependencies without dev-dependencies
-5. Sets current commit hash and plugin version in the plugin's main file
-6. Runs `wp dist-archive` to create the final archive (with builtin support for a `.distignore` file)
-7. Uploads it as an artifact for download or further processing
+4. Executes `rector` if configured by the package
+5. Re-installs dependencies without dev-dependencies
+6. Sets current commit hash and plugin version in the plugin's main file
+7. Runs `wp dist-archive` to create the final archive (with builtin support for a `.distignore` file)
+8. Uploads it as an artifact for download or further processing
```
tyrann0us, line 14: The base PR, #134, does not mention PHP-Scoper here. Let's do so in this PR according to the Boy Scout rules.
```suggestion
4. Executes PHP-Scoper if configured by the package
5. Executes Rector if configured by the package
6. Re-installs dependencies without dev-dependencies
7. Sets current commit hash and plugin version in the plugin's main file
8. Runs `wp dist-archive` to create the final archive (with builtin support for a `.distignore` file)
9. Uploads it as an artifact for download or further processing
```

tyrann0us (approved):
Thank you for working on this. While the changes LGTM code-wise, the addition of Rector to the reusable workflows can only be temporary until all products require modern PHP versions natively, IMO.
Left two minor comments that don't stop me from approving. Thanks!

tyrann0us (approved):
Thanks for incorporating the changes!

AlexP11223:
> the addition of Rector to the reusable workflows can only be temporary until all products require modern PHP versions natively, IMO.

I think products will never use only the latest version unless something greatly changes in WP/PHP world. We will probably still need to support PHP 8 when PHP 9 will be available because many merchants are slow to upgrade.

Also Rector is quite generic tool, it just transforms the code via AST using the specified rules, so maybe we find some other use cases for it.

PR #136, opened 2024-07-10
Add archive check workflow

AlexP11223:

 feature (new workflow)

The workflow runs some checks on the plugin archive to make sure that everything is ok (especially important if processing the source code via php-scoper or Rector).

It executes Parallel Lint to check the PHP syntax, and [Plugin Check](https://github.com/WordPress/plugin-check) phpcs ruleset, as well as other specified WP sniffs (by default `I18n`, because in some cases the `translators:` comments can be deleted by php-scoper).

It clones the Plugin Check repo and installs its composer deps, this seems to be the only quick way to use Plugin Check without WP and Docker, and it's a simple way to get `phpcs` and WP sniffs, since the plugin archive does not include the dev deps.

One issue here is that phpcs CLI is more limited than the config file. For example it’s not possible to ignore only the exception escaping errors when running the escaping sniff. But copying and maintaining the config for archive checking in each plugin seems like too much work, so for now trying like this.

### DIFF
--- a/.github/workflows/check-plugin-archive.yml
+++ b/.github/workflows/check-plugin-archive.yml
@@ -1,0 +1,80 @@
+name: Check plugin package
+
+on:
+  workflow_call:
+    inputs:
+      ARTIFACT:
+        description: The name of the generated artifact, usually the output of the archive creation workflow.
+        required: true
+        type: string
+      PLUGIN_FOLDER_NAME:
+        description: The name of the plugin folder (falls back to the repository name).
+        default: ''
+        required: false
+        type: string
+      EXCLUDED_SNIFFS:
+        description: The comma-separated list of excluded sniffs.
+        default: 'WordPress.WP.AlternativeFunctions'
+        required: false
+        type: string
+      ADDITIONAL_WP_SNIFFS:
+        description: The comma-separated list of additional sniffs to run.
+        default: 'WordPress.WP.I18n'
+        required: false
+        type: string
+      COMPOSER_ARGS:
+        description: The arguments passed to Composer when installing dependencies.
+        default: '--no-dev --prefer-dist'
+        required: false
+        type: string
+      PHP_VERSION:
+        description: The PHP version to use.
+        default: '7.4'
+        required: false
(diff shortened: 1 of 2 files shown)


luislard:
> One issue here is that phpcs CLI is more limited than the config file. For example it’s not possible to ignore only the exception escaping errors when running the escaping sniff. But copying and maintaining the config for archive checking in each plugin seems like too much work, so for now trying like this.

See [how we added a default config file here](https://github.com/inpsyde/reusable-workflows/blob/main/.github/workflows/automatic-release.yml#L88-L111),

.github/workflows/check-plugin-archive.yml
```diff
@@ -0,0 +1,80 @@
...
+        run: echo "plugin-folder-name=${PLUGIN_FOLDER_NAME:-${{ github.event.repository.name }}}" >> $GITHUB_OUTPUT
+
+      - name: Setup PHP with tools
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ inputs.PHP_VERSION }}
+          tools: parallel-lint
+
+      - name: Run Parallel Lint
+        run: parallel-lint .
```
Chrico, line 60: We already have a reusable workflow for this: https://github.com/inpsyde/reusable-workflows/blob/main/.github/workflows/lint-php.yml

Chrico:
@Biont and @AlexP11223 , what is the status of this PR?

Biont (changes requested):
I think the installation and usage of the Plugin Check package is worth revisiting. The current approach is pretty much a ship of theseus in that we 

- set up a generic php environment
- download the dev version of the plugin
- extract the rules we need
- execute them outside of the plugin

I think given that the plugin [ships with a ready-to-use wp-env](https://github.com/WordPress/plugin-check/tree/trunk#built-in-development-environment-optional), we could instead try to mount our plugin artifact into it and then run the official wp-cli command.

This would ensure maximum forward-compatibility in case they ever add more sophisticated check than just phpcs.

PR #147, opened 2024-08-06
Remove deprecated `BUILT_BRANCH_SUFFIX` input

### LINKED ISSUE #146: [Feature Request]: Remove `BUILT_BRANCH_SUFFIX`

The `BUILT_BRANCH_SUFFIX` was deprecated in #124 and consequently removed from all calling workflows.

Therefore, it is save to remove the input from the workflow and the documentation.

-

### Additional context

_No response_

### Code of Conduct

tyrann0us:

Feature
Closes #146 

The deprecated `BUILT_BRANCH_SUFFIX` input is present. All calling workflows have been updated to use the newer `BUILT_BRANCH_NAME`.

The `BUILT_BRANCH_SUFFIX` input is removed.

Yes, but no package uses the deprecated input anymore.

**Other information**:
I haven't tested this, so I'd appreciate it if one of you could give this a spin. Thanks!

### DIFF
--- a/.github/workflows/build-and-push-assets.yml
+++ b/.github/workflows/build-and-push-assets.yml
@@ -202,7 +202,7 @@ jobs:
       - name: Set branch environment variables
         if: ${{ github.ref_type == 'branch' }}
         run: |
-          echo "BUILT_BRANCH_NAME=${{ inputs.BUILT_BRANCH_NAME && inputs.BUILT_BRANCH_NAME }}" >> $GITHUB_ENV
+          echo "BUILT_BRANCH_NAME=${{ inputs.BUILT_BRANCH_NAME && inputs.BUILT_BRANCH_NAME || github.ref_name }}" >> $GITHUB_ENV
 
       - name: Set tag environment variables
         if: ${{ github.ref_type == 'tag' }}
--- a/.github/workflows/build-and-push-assets.yml
+++ b/.github/workflows/build-and-push-assets.yml
@@ -279,8 +279,11 @@ jobs:
           git add -A
           git commit -m "[BOT] Add compiled assets for #${{ github.ref }}" --no-verify || ((echo "NO_CHANGES=yes" >> $GITHUB_ENV) && (echo "No changes to commit"))
 
+#      The logic should be like the following:
+#      If we have changes in our assets we commit and push them.
+#      If we push not to the same branch, we always push commits (because they originated from the development branch).
       - name: Git push for branch
-        if: ${{ github.ref_type == 'branch' && (env.NO_CHANGES != 'yes' || inputs.BUILT_BRANCH_NAME != '') }}
+        if: ${{ github.ref_type == 'branch' && (env.NO_CHANGES != 'yes') }}
         run: git push
 
       - name: Git push for tag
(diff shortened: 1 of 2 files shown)


dottxado (approved):
Tested, it works 💪

.github/workflows/build-and-push-assets.yml
```diff
@@ -212,7 +202,7 @@ jobs:
       - name: Set branch environment variables
         if: ${{ github.ref_type == 'branch' }}
         run: |
-          echo "BUILT_BRANCH_NAME=${{ inputs.BUILT_BRANCH_NAME && inputs.BUILT_BRANCH_NAME || format('{0}{1}', github.ref_name, inputs.BUILT_BRANCH_SUFFIX) }}" >> $GITHUB_ENV
+          echo "BUILT_BRANCH_NAME=${{ inputs.BUILT_BRANCH_NAME && inputs.BUILT_BRANCH_NAME }}" >> $GITHUB_ENV
```
shvlv, line 205:
```suggestion
          echo "BUILT_BRANCH_NAME=${{ inputs.BUILT_BRANCH_NAME && inputs.BUILT_BRANCH_NAME || github.ref_name }}" >> $GITHUB_ENV
```

The thing is, we can't remove support for the legacy (or, to be more precise, an original) approach to push compiled to the **current** branch. If we really want this, we should make the `BUILT_BRANCH_NAME` input required. It should be prepended deprecation step:

```yaml
- name: Deprecation warning
        if: ${{ inputs.BUILT_BRANCH_NAME = '' }}
        run: |
          echo '::warning::The BUILT_BRANCH_NAME will be required soon. If you push compiled assets to the current, branch you should use ${{ github.ref_name }} value.'
``` 

But the main question is what we consider the upstream approach :smile:

shvlv (approved):

.github/workflows/build-and-push-assets.yml
```diff
@@ -290,7 +280,7 @@ jobs:
           git commit -m "[BOT] Add compiled assets for #${{ github.ref }}" --no-verify || ((echo "NO_CHANGES=yes" >> $GITHUB_ENV) && (echo "No changes to commit"))
 
       - name: Git push for branch
-        if: ${{ github.ref_type == 'branch' && (env.NO_CHANGES != 'yes' || inputs.BUILT_BRANCH_SUFFIX != '' || inputs.BUILT_BRANCH_NAME != '') }}
+        if: ${{ github.ref_type == 'branch' && (env.NO_CHANGES != 'yes' || inputs.BUILT_BRANCH_NAME != '') }}
```
luislard, line 283: Shouldn't `BUILT_BRANCH_NAME` be read from env since, in line 205, we fall back to `GitHub.ref_name`?

We might end in an infinite workflow run loop; see: 
Someone pushes to dev-main, and `BUILT_BRANCH_NAME` is not defined; the workflow runs and pushes the build to the same branch due to the assignment in 205.

Do we want to allow then an empty `BUILT_BRANCH_NAME` to continue with the workflow to push to the same branch?

```suggestion
        if: ${{ github.ref_type == 'branch' && (env.NO_CHANGES != 'yes' || env.BUILT_BRANCH_NAME != '') }}
```

luislard: we decided to leave it as it is, I did a test without the change and there is no infinite loop happening

Chrico (approved):

PR #158, opened 2024-09-10
Add actionlint workflow

### LINKED ISSUE #139: [Bug]: Check workflows
### Description of the bug

Workflows may contain a bug.

### Reproduction instructions

https://github.com/szepeviktor/reusable-workflows/actions/runs/9998829171/workflow

output:
https://github.com/szepeviktor/reusable-workflows/actions/runs/9998829171/job/27638576561

### Expected behavior

no errors in workflows

### Environment info

_No response_

### Relevant log output

_No response_

### Additional context

Please use Actionlint, fixing this single bug is only for today.

### Code of Conduct

tyrann0us:

Feature

Workflow files are not checked and may contain syntax errors.

All workflows files are checked with https://github.com/rhysd/actionlint.

**Other information**:
Fixes #139
This is not an internal workflow but should be consumable by calling workflows.

### DIFF
--- a/.github/workflows/lint-workflows.yml
+++ b/.github/workflows/lint-workflows.yml
@@ -22,4 +22,4 @@ jobs:
       - name: Run actionlint
         uses: docker://rhysd/actionlint:latest
         with:
-          args: -color -pyflakes=
+          args: -color -pyflakes= -shellcheck=
--- a/.github/workflows/lint-workflows.yml
+++ b/.github/workflows/lint-workflows.yml
@@ -3,10 +3,8 @@ name: Lint GitHub Actions workflows
 on:
   push:
     paths:
-      - '.github/workflows/*.yml'
+      - '**.yml'
   pull_request:
-    paths:
-      - '.github/workflows/*.yml'
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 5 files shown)


shvlv (commented):
Do I understand correctly that the workflow is bi-purposed: for our internal lint and for consuming as a reusable one? If so, should we add the documentation for consumers? Maybe there is not much to say, but still...

.github/workflows/lint-workflows.yml
```diff
@@ -0,0 +1,26 @@
+name: Lint GitHub Actions workflows
+
+on:
+  push:
+    paths:
+      - '**.yml'
```
dottxado, line 6: Today I stumbled upon one of our internal plugins that is using `.yml` files as configuration files, not for workflows. If we are going to use this workflow only to lint workflows, we can be more specific here on the path, while if we are going to use this also to lint our actions, maybe would it be cleaner to remove this, leaving the choice to the caller? :thinking:

shvlv: I think the caller will define their own value for `on` because when you use workflow, you redefine only the `jobs` part.

shvlv (approved):

dottxado (approved):

Chrico (approved):

szepeviktor:
Well done.

PR #170, opened 2024-12-12
Pre-install `sniccowp/php-scoper-wordpress-excludes`

tyrann0us:

Feature

The "Create plugin archive" [workflow](https://github.com/inpsyde/reusable-workflows/blob/main/.github/workflows/build-plugin-archive.yml), where [PHP-Scoper](https://github.com/humbug/php-scoper) is run if `scoper.inc.php` is present, runs `composer install --no-dev` (understandably). However, this requires the helper library [`sniccowp/php-scoper-wordpress-excludes`](https://github.com/snicco/php-scoper-wordpress-excludes) to be installed as a regular dependency to be available in the build process while it clearly is a dev "sub" dependency.

Since we already [install](https://github.com/inpsyde/reusable-workflows/blob/main/.github/workflows/build-plugin-archive.yml#L168) `humbug/php-scoper` with [`shivammathur/setup-php`](https://github.com/shivammathur/setup-php) in the release workflow and `sniccowp/php-scoper-wordpress-exclude` is a hard requirement for all our packages where we scope namespaces, we add it to the list of tools for `shivammathur/setup-php` to install. This allows us to remove the tool from the regular dependencies of packages using this workflow.

No. Existing packages requiring `sniccowp/php-scoper-wordpress-exclude` and referencing local `vendor/` paths to the list of excluded WordPress symbols (see documentation: https://github.com/humbug/php-scoper/blob/main/docs/further-reading.md#example-for-wordpress-core) continue to work.

**Other information**:
As part of the Boy Scout rule, I linked the tools used in the workflow in the documentation and removed the false claim that Composer Asset Compiler and WordPress Translation Downloader are installed when "required and configured by the package". They are installed when the workflow finds their configuration in the calling package, even when they are not _required_.

### DIFF
--- a/.github/workflows/build-plugin-archive.yml
+++ b/.github/workflows/build-plugin-archive.yml
@@ -165,7 +165,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ inputs.PHP_VERSION_BUILD }}
-          tools: humbug/php-scoper, rector
+          tools: humbug/php-scoper, sniccowp/php-scoper-wordpress-excludes, rector
 
       - name: Check optional Composer build tools
         id: composer-tools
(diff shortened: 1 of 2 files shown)


Biont (approved):
:heart:

dottxado (approved):

PR #178, opened 2025-02-04
Explain adjustments of composer file autoloading

Biont:

docs update

n/a

n/a

No

**Other information**:
There is a small customization about composer's file autoloading specific to this workflow. Coming up with this line of code actually involved quite a bit of exploration (and also the search of a better place to propose a fix).
Today, from some distance, it is probably hard to understand what this code really DOES, so I am intending to add more context inline.

For more context, check the comments on the initial PR:
https://github.com/inpsyde/reusable-workflows/pull/134/files#r1647580638

### DIFF
--- a/.github/workflows/build-plugin-archive.yml
+++ b/.github/workflows/build-plugin-archive.yml
@@ -239,11 +239,14 @@ jobs:
         run: |
           rector
 
-      - name: Run PHP-Scoper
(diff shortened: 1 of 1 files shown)


tyrann0us (approved):
We currently face the issue this step actually should prevent, so I'm unsure if there might be more changes coming, but for now, code-wise LGTM. Thanks!

PR #194, opened 2025-07-08
Unified build

### LINKED ISSUE #179: Replace `NO_CHANGES` with `HAS_GIT_CHANGES`
IMO, `HAS_GIT_CHANGES=no` is much better than `NO_CHANGES=yes`.
I aligned with others, to avoid changing others in this PR.

_Originally posted by @luislard in https://github.com/inpsyde/reusable-workflows/pull/174#discussion_r1942926935_

Biont:
- [x] Tests for the changes have been added (for bug fixes/features)

Feature - New reusable workflow for building and pushing plugin/theme archives to build branches

**What is the current behavior?**
Previously, we had a workflow for creating plugin archives that required manual artifact storage and management. The build process was more complex and didn't provide an automated way to access builds directly via GitHub branches.

This PR introduces a new `build-push.yml` reusable workflow that:

### Core Functionality
- **Automatic build branch determination**: Inspects the origin branch and creates corresponding build branches while respecting our branching strategy  `dev/` prefix (e.g., `dev/main` → `main`)
- **Intelligent version handling**: When `PACKAGE_VERSION` is not provided, automatically fetches the latest release and creates semver-compatible pre-release versions using the normalized branch name (e.g., `1.2.3-feature-user-auth`)
- **Comprehensive build process**: Handles both plugins and themes with proper version header updates, asset compilation, and dependency management

### Build Process Features
- Installs production dependencies via Composer with configurable arguments
- Compiles assets using `npm run build` with Node.js setup
- Updates version information in plugin/theme headers and `package.json`
- Integrates with WordPress Translation Downloader when configured
- Supports PHP-Scoper for dependency prefixing with proper autoloader handling
- Applies `.distignore` file filtering to exclude development files
- Commits and pushes build artifacts to determined build branches
- Creates GitHub Actions artifacts for download (excluding `.git` folder)

### Key Benefits
- **Direct GitHub access**: Build artifacts are accessible directly via build branches, eliminating need for external artifact storage
- **Automatic versioning**: Reduces manual version management overhead while maintaining semver compliance
- **Flexible configuration**: Supports extensive customization through inputs and secrets
- **Signed commits**: Supports SSH key-based commit signing for security
- **Clean artifacts**: Properly excludes `.git` folder and respects `.distignore` patterns

### Configuration Options
- Configurable Node.js and PHP versions
- Support for private npm registries
- Custom PHP tools installation
- Flexible Composer arguments
- Pre-script execution for custom build steps
- Override capability for build branch naming
- Comprehensive environment variable support

**Does this PR introduce a breaking change?**
No breaking changes to existing workflows. This is a new workflow that can be adopted independently. However, teams adopting this workflow should be aware of:

- **New branch naming convention**: Build artifacts will be pushed to `dev/{origin-branch}` branches
- **Version format changes**: Auto-generated versions follow `{latest-release}-{normalized-branch}` pattern
- **Required secrets**: Workflow requires specific GitHub user credentials for commit signing and pushing

**Other information**:

### Migration Path
Teams can gradually migrate from existing build processes by:
1. Adding the new workflow alongside existing ones
2. Testing with non-critical branches first
3. Updating deployment processes to reference new build branches
4. Removing old workflows once confidence is established

### Security Considerations
- All commits are signed using SSH keys for authenticity
- Supports private npm registries with token authentication
- Composer authentication for private packages
- Proper cleanup of temporary signing keys

### Documentation
- Comprehensive documentation added following existing patterns
- Examples for both simple and complex use cases
- Clear explanation of automatic version handling
- Best practices and recommendations included

### Testing Recommendations
- Test with different branch naming patterns
- Verify version generation with and without existing releases
- Confirm build artifacts exclude appropriate files
- Validate signed commits and build branch creation

This workflow significantly improves our CI/CD capabilities by providing a standardized, automated approach to building and distributing plugin/theme archives while maintaining flexibility for project-specific requirements.


.github/workflows/build-push.yml
```diff
@@ -0,0 +1,289 @@
...
+            NORMALIZED_BRANCH=$(echo "${{ github.ref_name }}" | sed 's/[^a-zA-Z0-9-]/-/g' | sed 's/--*/-/g' | sed 's/^-\|-$//g' | tr '[:upper:]' '[:lower:]')
+          
+            # Construct semver with pre-release identifier
+            PACKAGE_VERSION="${LATEST_RELEASE}-${NORMALIZED_BRANCH}"
+            echo "Generated package version: $PACKAGE_VERSION"
+          fi
+          
+          echo "PACKAGE_VERSION=$PACKAGE_VERSION" >> $GITHUB_ENV
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_USER_SSH_KEY && secrets.GITHUB_TOKEN || github.token }}
```
Chrico, line 163: can we put the `env` before `run` ?

.github/workflows/build-push.yml
```diff
@@ -0,0 +1,289 @@
...
+          script: |
+            JSON
+              .parse(process.env.ENV_VARS)
+              .forEach(envVar => core.exportVariable(envVar.name, envVar.value));
+
+      - name: Determine package version
+        run: |
+          if [ -n "${{ inputs.PACKAGE_VERSION }}" ]; then
+            PACKAGE_VERSION="${{ inputs.PACKAGE_VERSION }}"
+            echo "Using provided package version: $PACKAGE_VERSION"
```
Chrico, line 140: Is this "if" required? The `env.PACKAGE_VERSION` is already set above the steps. So maybe we can run this step with an `if:`-condition?

.github/workflows/build-push.yml
```diff
@@ -0,0 +1,289 @@
...
+        with:
+          node-version: ${{ inputs.NODE_VERSION }}
+          registry-url: ${{ inputs.NPM_REGISTRY_DOMAIN }}
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Set plugin version header
+        if: ${{ hashFiles('style.css') == '' }}
```
Chrico, line 213: `style.css` is part of WordPress Themes. Do we need to adjust the `if:`-condition here to `index.php` or do we need to update the `name:` to "Theme"? 👀

.github/workflows/build-push.yml
```diff
@@ -0,0 +1,289 @@
...
+        run: |
+          sed -Ei "s/Version: .*/Version: ${{ env.PACKAGE_VERSION }}/g" $(grep -rl --include "*.php" "Plugin Name:" . | head -1)
+
+      - name: Add commit hash to plugin header
+        if: ${{ hashFiles('style.css') == '' }}
```
Chrico, line 218: Same question as https://github.com/inpsyde/reusable-workflows/pull/194/files#r2194270750

.github/workflows/build-push.yml
```diff
@@ -0,0 +1,289 @@
...
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_REGISTRY_TOKEN }}
+        with:
+          node-version: ${{ inputs.NODE_VERSION }}
+          registry-url: ${{ inputs.NPM_REGISTRY_DOMAIN }}
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Set plugin version header
```
Chrico, line 230: I'm adding this to the top of the whole section. Shoudl we maybe in general combine "Prepare Plugin" and "Prepare Theme" as 2 steps in total instead of having multiple steps and checks for version, commit hash, ...?

.github/workflows/build-push.yml
Biont: Yeah we've been tripping over this before. I think we can afford a variable that clearly tells us what this is about

.github/workflows/build-push.yml
```diff
@@ -0,0 +1,289 @@
...
+  build-push:
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
+    env:
+      NO_CHANGES: ''                                   # we'll override if no changes to commit
+      # Disables symlinking of local "path" repositories during package installation.
+      # During development, symlinking is preferable.
+      # In production builds, we ship the actual install location and remove the source directory.
+      COMPOSER_MIRROR_PATH_REPOS: 1
+      GIT_SHA: ${{ github.sha }}
```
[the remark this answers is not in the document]
Biont: Required not. It could theoretically be cleaner to access in shell scripts, but we are not even doing that (still using `${{ ... }}` syntax ), so I guess we can change this

.github/workflows/build-push.yml
```diff
@@ -0,0 +1,289 @@
...
+          git checkout ${{ github.ref_name }} -- .
+
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ inputs.PHP_VERSION }}
+          tools: php-scoper, sniccowp/php-scoper-wordpress-excludes, composer/installers, inpsyde/wp-translation-downloader, ${{ inputs.PHP_TOOLS }}
+          coverage: none
+
+      - name: Install Composer dependencies without dev dependencies
```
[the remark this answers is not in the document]
Biont: This is unchanged from the existing `build-plugin-archive.yml` workflow. I actually believe there even was a time when the `--no-dev` was hardcoded next to other args coming from the input.

But this also harkens back to a time when we needed to check out composer twice:
Once for getting all the tooling (asset compiler, translation downloader etc) and once for getting the "production-ready" state. This is no longer needed. 

So today, I think this is mostly about the wording. There isn't actually a need to spell out the "without dev dependencies" here any longer, so I'm just going to remove it. The input/config is fine in my book, but I am also not married to it. For now I'll just change the comment because it is the smaller change. Let me know if this works for you.

tyrann0us (commented):
Thank you so much for working on this! This could be a real game-changer and massively improve cross-team collaboration with a unified asset compilation. No more learning about other workflows or strange variations. Thank you for that! 💪🏽
I finally took the time to review it because it will be OOF for the next two weeks, and I didn't want to block this any longer. I left several mostly minor comments and suggestions.

.github/workflows/build-push.yml
```diff
@@ -0,0 +1,314 @@
...
+      PHP_VERSION_BUILD:
+        description: PHP version to use when executing build tools.
+        default: '8.2'
+        required: false
+        type: string
+      PACKAGE_FOLDER_NAME:
+        description: The name of the package folder (falls back to the repository name).
+        required: false
+        default: ''
+        type: string
```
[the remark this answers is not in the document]
Biont: Okay this is a bit unfortunate. There are several angles here

- **Some plugins need to be able to customize the plugin folder** so that the _install location_ can be chosen independently from the repo/archive name. Usually this has to do with maintaining compatibility with an existing plugin slug on wp-org SVN - and maintaining a frictionless (auto-)update path for existing plugin users. But scenarios are also imaginable where you need to stay consistent with an existing WordPress installation despite extracting a bunch of plugins into a `<CLIENT>-awesome-plugin` repo naming scheme

- **We've been asserting that plugin archives do not need to contain a subfolder to be installable**. The existing workflow (or rather: the  **wp dist-archive** command ) [does this](https://github.com/inpsyde/reusable-workflows/blob/a810bb31acd37bfbf209c5da79e65b9ade003495/.github/workflows/build-plugin-archive.yml#L279) - and here we've previously been placing the folder customization. But since it's not a technical requirement and not commonly used in agency projects that use build branches, we decided to approach this with a KISS philosophy.

- **Build branches with only a folder in it look pretty bad**. Yes this is purely aesthetics and I do not expect developers to often browse these branches, but I remember us discussing that it looks less confusing without.

- We frequently hand over individual feature packages to QA. So a way to make sense of individual archives - as well as maintaining a consistent install folder name - could be a requirement.
---

The new workflow - as of now - would result in an install folder like this if you install from a zip file (either from artifact or manual GutHub zip download):

screenshot_2025-07-29-090009

This is because we combine `PACKAGE_FOLDER_NAME` and `PACKAGE_VERSION`. For manual source code downloads, it combines the repo name with the current tree/branch. 

Whereas when I put it all into a dedicated subfolder, it looks like this:
screenshot_2025-07-29-090938

I guess WP falls back to the zip filename if no subfolder is present.

---

So all things considered I feel there is no way to avoid a subfolder inside the build branches. Are there any experiences with the composer installer workflows that would speak against this? That's the one thing I did not check

docs/build-push.md
```diff
@@ -0,0 +1,255 @@
...
+4. Updates version information in plugin/theme headers and `package.json`
+5. Executes [WordPress Translation Downloader](https://github.com/inpsyde/wp-translation-downloader) if configured by the package
+6. Executes [PHP-Scoper](https://github.com/humbug/php-scoper) if configured by the package
+7. Applies `.distignore` file filtering if present
+8. Commits and pushes the build artifact to the determined build branch
+9. Uploads the build as a GitHub Actions artifact for download
+
+## Branch naming convention
+
+The workflow implements an inverted branch scheme where development branches are prefixed with `dev/` and build branches contain the compiled assets:
```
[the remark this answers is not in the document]
Biont: Because it reserves the "top-level" branch segment for release artifacts and thereby turns the actual source code into a second-class citizen `main (artifact) <-- dev/main (source)`.

All popular git flows I'm aware of would keep source code at the top, then use `feature/foo-bar` and `release/1.2.3` as a convention for separate purposes.

It would follow to use something like `build/1.2.3` for release artifacts if we're looking at this purely through the lens of git branching strategies. 

If I understood correctly, the setup we chose is mainly because it's more convenient for composer - ie. `composer require syde/acme:dev-main` instead of `composer require syde/acme:dev-build/main`.

.github/workflows/build-push.yml
```diff
@@ -0,0 +1,341 @@
...
+              .forEach(envVar => core.exportVariable(envVar.name, envVar.value));
+
+
+      - name: Determine package version
+        if: ${{ !inputs.PACKAGE_VERSION }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_USER_SSH_KEY && secrets.GITHUB_TOKEN || github.token }}
+        run: |
+          # Fetch latest public release
+          LATEST_RELEASE=$(gh release list --limit 1 --exclude-drafts --exclude-pre-releases --json tagName --jq '.[0].tagName' 2>/dev/null || echo "")
```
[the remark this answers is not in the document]
Biont: I researched an alternative approach that 

1. Works with git instead of gh
2. Does not rely on releases - works purely with tags
3. Is quite a bit faster (I assume the gh cli does some API calls here)

The theoretical downside is that **tags** are not **releases**: You can create tags for completely arbitrary reasons ( `before-giant-refactor`, `last-php7-compatible` etc - purely in order to label commits as human-friendly points in time you may wish to jump back to ). A GH Release -while being backed by tags carries at least some credible expectation of being tied to a (semantic) version number.

So this workflow might run into problems in repos that use both sem-ver releases as well as arbitrary tags: It would simply pick the latest tag instead of the latest release in these cases and yield undesired results.

But I guess this is an edge case - and we can still use `inputs.PACKAGE_VERSION` in the calling workflows to specify the latest release via the gh cli.

.github/workflows/build-push.yml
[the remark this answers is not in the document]
Chrico: In this case, when outputting data with `<type>: <value>`, no full stop is required. But in case of full sentences, we should add a full stop.

Example: https://github.com/inpsyde/reusable-workflows/blob/de379f96149671a6196f16f469ad77df5b3431ed/.github/workflows/build-push.yml#L140

> No releases found, using default version 0.0.0.

.github/workflows/build-push.yml
```diff
@@ -0,0 +1,335 @@
...
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Determine package type
+        run: |
+          if [ -f "style.css" ]; then
+            echo "Package type: WordPress Theme"
+            echo "IS_THEME=true" >> $GITHUB_ENV
+          else
+            echo "Package type: WordPress Plugin"
```
Chrico, line 253: What about Libraries?

.github/workflows/build-push.yml
```diff
@@ -0,0 +1,341 @@
...
+            echo "IS_THEME=false" >> $GITHUB_ENV
+          fi
+
+      - name: Prepare Plugin
+        if: ${{ env.IS_THEME == 'false' }}
+        run: |
+          # Find and update plugin main file
+          PLUGIN_FILE=$(grep -rl --include "*.php" "Plugin Name:" . | head -1)
```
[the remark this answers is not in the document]
Chrico: > I am certain we did not. Not only would that prevent existing plugin projects from adopting the workflow, it is simply not a convention that exists outside of our company. The official developer guidelines explicitly ask for the file to be named after the containing folder.

We should align here. While we state in a few areas "this is how you have to use this workflow because _we_ decided it", here we go the other way around and state "outside of our company".

.github/workflows/build-push.yml
```diff
@@ -0,0 +1,335 @@
...
+          if [ -n "$PLUGIN_FILE" ]; then
+            echo "Updating plugin file: $PLUGIN_FILE"
+            sed -Ei "s/Version: .*/Version: ${{ env.PACKAGE_VERSION }}/g" "$PLUGIN_FILE"
+            sed -Ei "s/SHA: .*/SHA: ${{ github.sha }}/g" "$PLUGIN_FILE"
+          else
+            echo "Warning: No plugin main file found!"
+          fi
+
+      - name: Prepare Theme
```
Chrico, line 270: While I think it is good to have a check for the Plugins Headers and Theme style.css Headers, should we maybe also support something dynamic like a global constant, which could also contain the version? This could become handy for libraries that do not have a version set anywhere in code and are parseable through WordPress.

.github/workflows/build-push.yml
Chrico: I suppose it's just a matter of wording. :) We have a few libraries that also ship JS/CSS and need the build workflow. 💪🏻

.github/workflows/build-push.yml
Chrico: We have some libraries (not Plugins) which also contain a version that is shown in the Backend, or, depending on the version, something needs to be triggered (like updating the DB tables). In this case, a version string also needs to be added and updated manually somewhere in a file. 
Through `inpsyde/modularity`, we only have the possibility right now to set `version` in the `composer.json` for libraries - see https://github.com/inpsyde/modularity/blob/master/src/Properties/LibraryProperties.php#L89-L93

My thought here was to also cover the need to have a "version" changed in code for "PHP libraries".

Biont: https://github.com/inpsyde/reusable-workflows/pull/194/commits/91021336700af3e9011e019bf2181dd883d2289b

.github/workflows/build-push.yml
Biont: I refactored things a little, so that we are only checking against a generic `PACKAGE_TYPE` from now on. This would make it easier to address types other than plugins/themes in the future.
https://github.com/inpsyde/reusable-workflows/pull/194/commits/d0deb15ae4ec6a34a6afd8cc847d958960ae1ee2

.github/workflows/build-push.yml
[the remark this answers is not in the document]
Biont: https://github.com/inpsyde/reusable-workflows/pull/194/commits/5f04664ecf828d015f648a845a3e5918ec2abfe1 introduces a subfolder within archives

.github/workflows/build-push.yml
```diff
@@ -0,0 +1,341 @@
...
+            echo "Latest release found: $LATEST_RELEASE"
+          fi
+          
+          # Strip 'dev/' prefix from branch name for version string
+          ORIGIN_BRANCH="${{ github.ref_name }}"
+          if [[ "$ORIGIN_BRANCH" == dev/* ]]; then
+            VERSION_BRANCH="${ORIGIN_BRANCH#dev/}"
+          else
+            VERSION_BRANCH="$ORIGIN_BRANCH"
+          fi
```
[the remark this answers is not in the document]
Biont: https://github.com/inpsyde/reusable-workflows/pull/194/commits/5abc920b4ac5d596fbe974f96b7109c496bdcb7e

.github/workflows/build-push.yml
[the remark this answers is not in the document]
Biont: https://github.com/inpsyde/reusable-workflows/pull/194/commits/044964f1bc1b8ca8901fb0a93d8140837b933442

.github/workflows/build-push.yml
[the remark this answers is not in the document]
Biont: https://github.com/inpsyde/reusable-workflows/pull/194/commits/6eace1721b425a73b9c164b83e6caebbce8a4f9f

.github/workflows/build-push.yml
```diff
@@ -0,0 +1,366 @@
...
+        required: true
+      ENV_VARS:
+        description: Additional environment variables as a JSON formatted object.
+        required: false
+
+jobs:
+  build-push:
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
+    env: # Used to check whether there are changes to commit
```
Chrico, line 83: Just a small note, but can we keep comments consistently always above the line it is affecting? 

```yml
env: 
	# Used to check whether there are changes to commit
	HAS_GIT_CHANGES: ''
	# Disables symlinking of local "path" repositories during package installation.
	# During development, symlinking is preferable.
	# In production builds, we ship the actual install location and remove the source directory.
	COMPOSER_MIRROR_PATH_REPOS: 1
	# Will be overridden by determine package version step
	PACKAGE_VERSION: ${{ inputs.PACKAGE_VERSION }}
	PACKAGE_NAME: ${{ inputs.PACKAGE_NAME && inputs.PACKAGE_NAME || github.event.repository.name }}
```

tyrann0us:
Deprecation warning needs to be added to the "old" workflow and documentation.

tyrann0us (commented):
Thank you for implementing all the changes! Left a final round of very minor comments.

Apart from that, could you please add a deprecation warning to the "old" "build and push" workflow and documentation? Thank you!

tyrann0us (approved):
> Apart from that, could you please add a deprecation warning to the "old" "build and push" workflow and documentation? Thank you!

I added the deprecation warnings. @Chrico FYI, this now officially deprecates three workflows:
1. "Build and push assets"
2. "Assets compilation" (with Composer Assets Compiler)
3. "Create plugin archive"

Are you fine with that?

docs/build-push.md
```diff
@@ -0,0 +1,216 @@
...
+      CUSTOM_PACKAGE_VERSION:
+        description: 'Custom Package Version (skip auto-detection of)'
+        required: false
+      CUSTOM_BUILD_BRANCH:
+        description: 'Custom build branch name'
+        required: false
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
Chrico: Hard to say. If we have 2 PR's merged, only the last one that was merged matters, because this one also contains the changes from the first, right? The problem I see here is that we have merged the following:

- PR 1 -> merge
- PR 2 -> merge after PR 1 and has changes from 1 because 1 is "latest state" in "dev/master".

But if, for whatever reason, the build process for PR 1 takes longer than PR 2, then we will have as a result in the built repo "PR 1" pushed. 👀

tyrann0us (approved):
Thank you for continuously implementing all the changes! From my side, this is good to be merged. @dottxado, @Chrico, any last wishes? 😆

Chrico (approved):

PR #216, opened 2026-01-23
Prevent silent failures in test-playwright workflow

zhyian:

Improvement

The test step uses `continue-on-error` to allow artifact upload for debugging(my guess), but the workflow never checked the actual test outcome afterward. This caused test failures to be silently ignored, showing green CI status.

Add step ID and outcome check to properly fail the job when tests fail.

No


dottxado (approved):
Thanks @zhyian! It was a feature we needed to add :muscle:

Chrico (approved):

PR #231, opened 2026-03-06
Support custom Composer vendor directory locations

zhyian:

Feature

The workflows hardcode `./vendor/bin/` as the path to the binary files. Projects that configure a custom `vendor-dir` in `composer.json` (e.g. `public/wp-content/vendor/`) cannot use these workflows.

Tools are now executed using the binary path resolved dynamically via `composer config bin-dir`, instead of the hardcoded `./vendor/bin/`. This means workflows will automatically respect any custom `bin-dir` configured in the project's `composer.json`, supporting setups where the vendor directory is not in the repository root (e.g. `public/wp-content/vendor`). 

No

### DIFF
--- a/.github/workflows/coding-standards-php.yml
+++ b/.github/workflows/coding-standards-php.yml
@@ -28,6 +28,11 @@ on:
         default: '--graceful-warnings ./phpcs-report.xml'
         required: false
         type: string
+      VENDOR_DIR:
+        description: Path to the Composer vendor directory, relative to the repository root.
+        default: 'vendor'
+        required: false
+        type: string
     secrets:
       COMPOSER_AUTH_JSON:
         description: Authentication for privately hosted packages and repositories as a JSON formatted object.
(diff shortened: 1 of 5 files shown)


tyrann0us (commented):
**issue:**

Thank you for the PR! As it is right now, it is not merge-ready because all workflows that execute Composer libraries would need to be updated, and so would the documentation.

Anyway, I would like to request a different approach by having Composer do the `vendor`, or rather `bin` directory resolving:
```yml
- name: Run PHP_CodeSniffer
  run: ./$(composer config bin-dir)/phpcs ${{ inputs.PHPCS_ARGS }}
```

This has multiple benefits over the current approach, IMO:
* Does not require any changes to calling workflows.
* Does not require any documentation updates.
* DRY; no repetition of paths in `composer.json` and workflow files.
* Uses [`bin-dir`](https://getcomposer.org/doc/06-config.md#bin-dir) to support `composer.json` configs where `vendor-dir` and/or `bin-dir` is set (falls back to whatever `vendor` resolves to + `/bin`).

Please test this approach and apply it to all corresponding workflow files if it works. Thank you!

tyrann0us (commented):
> Does not require any documentation updates.

I'm sorry, this was a lie. 🙈  I realized that in https://github.com/inpsyde/reusable-workflows/blob/main/docs/php.md, we state multiple times, "It does so by executing the binary in the `./vendor/bin/` folder." This is potentially no longer accurate, and I would simply remove all occurrences of that sentence.

Also, can you apply the change to the new Deployer workflow https://github.com/inpsyde/reusable-workflows/blob/main/.github/workflows/deploy-deployer.yml added in #227?

(I know this is still a draft PR. 🙂)

tyrann0us (approved):
Thank you for working on this! 💪🏽 PR title and description no longer match the latest state, but this doesn't stop me from approving. 😉

Chrico (approved):

PR #234, opened 2026-03-20
feat(build-and-distribute): add NPM_BUILD_SCRIPT input to allow custom build scripts

luislard:
## Summary

- Add optional `NPM_BUILD_SCRIPT` input to `build-and-distribute` workflow (defaults to empty, falls back to `build`)
- Update the "Compile assets" step to use the input with `|| 'build'` fallback
- Document the new input and provide a conditional usage example in docs

## Usage example

```yml
with:
  NPM_BUILD_SCRIPT: ${{ (github.ref_name == 'dev/main' || github.ref_name == 'dev/beta' || github.ref_type == 'tag') && 'build' || 'build:dev' }}
```

## Test plan

- [ ] Verify workflow runs `npm run build` when `NPM_BUILD_SCRIPT` is not provided
- [ ] Verify workflow runs the specified script when `NPM_BUILD_SCRIPT` is set

Closes #207

### DIFF
--- a/.github/workflows/build-and-distribute.yml
+++ b/.github/workflows/build-and-distribute.yml
@@ -13,6 +13,11 @@ on:
         default: 18
         required: false
         type: string
+      NPM_BUILD_SCRIPT:
+        description: The script to be used to run the build. e.g. build:dev
+        default: ''
+        required: false
+        type: string
       NPM_REGISTRY_DOMAIN:
         description: Domain of the private npm registry.
         default: https://npm.pkg.github.com/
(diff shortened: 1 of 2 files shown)


luislard:
Quick question, if we go in the `MODE` direction:

Let's say we set `MODE: dev` 
Is it desired to also affect the composer installation? because now the composer args are default to `--no-dev`. 

There may be a caveat: the cache will need to be handled differently because we now have two build types, dev and prod.

If the `MODE` only affects the assets build, I think it is easier to keep it with only `NPM_BUILD_SCRIPT`. Otherwise, we would have to declare more inputs: 

```
MODE: dev
COMPILE_SCRIPT_DEV: 'build:dev'
COMPILE_SCRIPT_PRD: 'build'
```

With the `NPM_BUILD_SCRIPT`

```
NPM_BUILD_SCRIPT:   ${{ (github.ref_name == 'dev/main' || github.ref_name == 'dev/beta' || github.ref_type == 'tag') && 'build' || 'build:dev' }}
```

PR #235, opened 2026-03-20
feat(deploy-deployer): expose Deployer exit code and reason as workfl…

luislard:

Feature — exposes Deployer's process exit code and a parsed reason message as `workflow_call` outputs.

Closes #230

Deployer stdout (info, warnings, errors) is only visible in CI logs. Callers have no way to pass the deployment result to downstream jobs (e.g. Slack notifications).

Two new `workflow_call` outputs are available:

- `deploy_exit_code` — Deployer's process exit code (`0` = success, `1` = error, `2` = warning)
- `deploy_reason` — a human-readable message parsed from Deployer's output

Example usage in a calling workflow:

```yaml
slack-notification:
  needs: build-and-deploy
  if: always() && needs.build-and-deploy.result != 'skipped'
  steps:
    - uses: slackapi/slack-github-action@v2
      with:
        payload: |
          text: "Deploy result: ${{ needs.build-and-deploy.outputs.deploy_reason }}"
```

To make the outputs available even when the deployment fails, the workflow is split into two jobs:

- **`deploy`**: runs the full deployment and captures exit code + reason; always exits `0` so its outputs are committed
- **`assert-result`**: fails strictly on any non-zero exit code (including warnings, exit `2`), surfacing the reason via `::error::` annotation

This is necessary because GitHub Actions only exposes `workflow_call` outputs from jobs that complete successfully.

No breaking change in workflow behavior — the overall `workflow_call` result is still `failure` when Deployer fails, since `assert-result` fails.

There is a **visual change**: the `deploy` job will always appear green in the GitHub Actions UI. The failure is now visible in the `assert-result` job instead.

## Other information

N/A

### DIFF
--- a/.github/workflows/deploy-deployer.yml
+++ b/.github/workflows/deploy-deployer.yml
@@ -59,12 +59,22 @@ on:
       NPM_REGISTRY_TOKEN:
         description: Authentication for the private npm registry.
         required: false
+    outputs:
+      deploy_exit_code:
+        description: Deployer process exit code (0=success, 1=error, 2=warning).
+        value: ${{ jobs.deploy.outputs.exit_code }}
+      deploy_reason:
+        description: Human-readable deployment result message parsed from Deployer output.
+        value: ${{ jobs.deploy.outputs.reason }}
 jobs:
   deploy:
     runs-on: ubuntu-latest
     env:
       NODE_CACHE_MODE: ''
       COMPOSER_NO_DEV: 1
+    outputs:
+      exit_code: ${{ steps.run-deployer.outputs.exit_code }}
+      reason: ${{ steps.run-deployer.outputs.reason }}
 
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


Chrico:
@luislard, what is the status of this issue? Can we expect this to move forward and be ready for review and merge? :-)

.github/workflows/deploy-deployer.yml
```diff
@@ -160,11 +170,37 @@ jobs:
...
+          echo "reason=${reason}" >> $GITHUB_OUTPUT
+
+  assert-result:
+    needs: deploy
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fail on deployment error or warning
+        if: needs.deploy.outputs.exit_code != '0'
+        run: |
+          echo "::error::Deployer exited with code ${{ needs.deploy.outputs.exit_code }}: ${{ needs.deploy.outputs.reason }}"
```
dottxado, line 205: **issue**

This expression can expose security risks if any dependency in the chain is attacked and leaks commands into the `reason` string.
You should list the outputs in environment variables at the step level and let bash expand them, like 

```yml
assert-result:
    needs: deploy
    runs-on: ubuntu-latest
    steps:
      - name: Fail on deployment error or warning
        if: needs.deploy.outputs.exit_code != '0'
        env:
          EXIT_CODE: ${{ needs.deploy.outputs.exit_code }}
          REASON: ${{ needs.deploy.outputs.reason }}
        run: |
          echo "::error::Deployer exited with code ${EXIT_CODE}: ${REASON}"
          exit "${EXIT_CODE}"
```

dottxado (changes requested):
Hello @luislard, thanks for working on this 💪 
I have two requests/questions.

First, can you please update the documentation about the outputs? 🙇‍♀️ 

Then, why do we need two separate jobs? As you point out, this leads to the deployment job to always end green, and I don't really like it. You could merge the two jobs and add `continue-on-error: true` on the last step. If you can implement this change, ignore my comment down here 👇

PR #245, opened 2026-05-06
Fix auto-detection of latest package version

stracker-phil:

Bug fix

The automatically determined `PACKAGE_VERSION` defaulted to the most recently created git tag, regardless of whether that tag represented an actual release version or something else (e.g. arbitrary non-semver tags such as `archive/qa-legacy-ui-tests`).

This caused two problems downstream, where `PACKAGE_VERSION` is consumed:

- Build breaks on tags containing slashes. The "Prepare WordPress Plugin" and "Prepare WordPress Theme" steps interpolate `PACKAGE_VERSION` into a `sed` expression that uses `/` as its delimiter; a slash in the tag breaks the substitution with an error.
- Invalid versions without a slash silently land in project files. Non-semver tags get written into `package.json` (via `npm version`, which rejects them) and `composer.json` (via `jq`, which accepts any string), corrupting the build artifact.

The workflow now fetches the tag name of an actual GitHub Release via `gh release list`, so only published releases are considered.

Assumption: A release tag represents a **valid semantic version** and never contains slashes, e.g.

- `4.0.3`
- `4.0.3-alpha-2026-05-06`
- `v2.17.0`

The "0.0.0" fallback already existed previously, but it now triggers in more cases: any repository that has tags but no published GitHub Release will fall back to "0.0.0" instead of picking up a tag.

This will not block the build, but it will insert a "0.0.0"-based version string into project files in the following steps:

- Prepare WordPress Plugin
- Prepare WordPress Theme
- Update version in `package.json`
- Update version in `composer.json`

## Other information

Repositories that previously relied on tags being picked up without ever cutting a GitHub Release should publish a Release to keep getting a meaningful version prefix.

### `npm version` requirements

The workflow [constructs `PACKAGE_VERSION`](https://github.com/inpsyde/reusable-workflows/blob/37fbc05c85d592af4f24088a301a9ab8c7647b00/.github/workflows/build-and-distribute.yml#L164) and validates it against semver via`npm version` in the "[Update version in package.json](https://github.com/inpsyde/reusable-workflows/blob/37fbc05c85d592af4f24088a301a9ab8c7647b00/.github/workflows/build-and-distribute.yml#L318)" step. This means, the `LATEST_RELEASE` string itself must match the following pattern to pass this check:

```
[v]MAJOR.MINOR.PATCH[-IDENTIFIER]
```

Valid samples:
- `1.2.3`
- `v1.2.3`
- `1.2.3-alpha`
- `1.2.3-feature.1`
- `1.2.3-2026-05-21`

Invalid samples:
- `1.2`           → missing PATCH
- `1.2.3.4`     → too many segments
- `1.2.3-`       → empty identifier
- `v.1.2.3`      → dot after v
- `a.b.c`        → non-numeric segments

### DIFF
--- a/.github/workflows/build-and-distribute.yml
+++ b/.github/workflows/build-and-distribute.yml
@@ -139,7 +139,10 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_USER_SSH_KEY }}
         run: |
           # Fetch latest public release
-          LATEST_RELEASE=$(git for-each-ref --sort=-creatordate --count=1 --format='%(refname:short)' 'refs/tags/[0-9]*' || echo "")
+          LATEST_RELEASE=$(git for-each-ref --sort=-creatordate --format='%(refname:short)' refs/tags \
+            | grep -E '^v?[0-9]' \
+            | grep -v '/' \
+            | head -1 || echo "")
           
           if [ -z "$LATEST_RELEASE" ]; then
             echo "No releases found, using default version 0.0.0."
--- a/.github/workflows/build-and-distribute.yml
+++ b/.github/workflows/build-and-distribute.yml
@@ -141,14 +141,6 @@ jobs:
           # Fetch latest public release
           LATEST_RELEASE=$(gh release list --limit 1 --json tagName --jq '.[0].tagName' || echo "")
           
-          # Fallback: Fetch latest version tag (supports 1.2.3 and v1.2.3, ignores tags with slashes)
-          if [[ -z $LATEST_RELEASE ]]; then
-            LATEST_RELEASE=$(git for-each-ref --sort=-creatordate --format='%(refname:short)' refs/tags \
-              | grep -E '^v?[0-9]' \
-              | grep -v '/' \
-              | head -1 || echo "")
-          fi
-          
           if [ -z "$LATEST_RELEASE" ]; then
             echo "No releases found, using default version 0.0.0."
             LATEST_RELEASE="0.0.0"
--- a/.github/workflows/build-and-distribute.yml
+++ b/.github/workflows/build-and-distribute.yml
@@ -139,7 +139,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_USER_SSH_KEY }}
         run: |
           # Fetch latest public release
-          LATEST_RELEASE=$(git for-each-ref --sort=-creatordate --count=1 --format='%(refname:short)' refs/tags || echo "")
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-distribute.yml
```diff
@@ -139,7 +139,15 @@ jobs:
...
-          LATEST_RELEASE=$(git for-each-ref --sort=-creatordate --count=1 --format='%(refname:short)' refs/tags || echo "")
+          LATEST_RELEASE=$(gh release list --limit 1 --json tagName --jq '.[0].tagName' || echo "")
+          
+          # Fallback: Fetch latest version tag (supports 1.2.3 and v1.2.3, ignores tags with slashes)
+          if [[ -z $LATEST_RELEASE ]]; then
+            LATEST_RELEASE=$(git for-each-ref --sort=-creatordate --format='%(refname:short)' refs/tags \
+              | grep -E '^v?[0-9]' \
+              | grep -v '/' \
+              | head -1 || echo "")
+          fi
```
tyrann0us: I wonder whether we need that fallback. If the `gh` command fails, something is wrong at a broader level, so we should not continue creating the release anyway. 😉

tyrann0us (commented):
Please update the documentation, which still states that we are fetching the last _tag_ from the repository, whereas it is now the last _release_.

.github/workflows/build-and-distribute.yml
Biont: I agree with @tyrann0us 
The potential use would be to allow versioning via tags before any public release - purely via tags. But:

1. We already have a fallback to `0.0.0`
2. We can override the `PACKAGE_VERSION` if there is a strong reason to use the workflow in such a setup

In the other error scenario - `gh` failing - I also think it is wiser to stop and make developers/admins take a look

Biont (changes requested):
Thanks for this. I believe this was the correct approach all the time. I was "_emotionally_" in favor of a git-only (no gh cli) approach when I built this, but given that we are in the context of GitHub Actions, there's just no reason not to commit to their tooling.

IMO we can remove the fallback as per https://github.com/inpsyde/reusable-workflows/pull/245/changes#r3196373868 - then we are golden

.github/workflows/build-and-distribute.yml
stracker-phil: As @Biont suspected, the only reason for keeping this fallback was the case where a repo has no release yet (not because `gh` is unavailable). But the static `0.0.0` is possibly correct/sufficient for those project stages.

→ will remove the fallback, as I agree it's not adding real value

tyrann0us (approved):
Thank you for working on this! LGTM

Biont (approved):

PR #249, opened 2026-05-26
Add Codecov support to PHPUnit and Jest workflow

### LINKED ISSUE #188: [Feature Request]: tests-unit-php.yml does not support codecov

We should add to https://github.com/inpsyde/reusable-workflows/blob/main/.github/workflows/tests-unit-php.yml support for Codecov. 

The additional step could be like:

```yaml
            -   name: Coverage
                if: ${{ hashFile('coverage.xml') == '' }}
                uses: codecov/codecov-action@v4
                with:
                    token: ${{ secrets.CODECOV_TOKEN }}
                    files: ./coverage.xml
                    flags: unittests
                    verbose: true
```

-

### Additional context

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

tyrann0us:

Feature

The PHPUnit and Jest workflows don't support (conditionally) uploading coverage reports to Codecov.

Add Codecov support for the "Unit tests PHP" and "Unit tests JavaScript" workflows by uploading the respective coverage files to Codecov.

No. The tests are only run with coverage support, i.e., write to a coverage report file, when the `CODECOV_TOKEN` secret is set. And only if these files exist (`coverage/lcov.info` for Jest and `coverage.xml` for PHPUnit).

## Other information
Closes #188

### DIFF
--- a/.github/workflows/tests-unit-php.yml
+++ b/.github/workflows/tests-unit-php.yml
@@ -23,12 +23,20 @@ on:
         default: '--coverage-text'
         required: false
         type: string
+      CODECOV_COVERAGE_FILE:
+        description: Path to the coverage file to upload to Codecov. Upload is skipped when the file does not exist.
+        default: 'coverage.xml'
+        required: false
+        type: string
     secrets:
       COMPOSER_AUTH_JSON:
         description: Authentication for privately hosted packages and repositories as a JSON formatted object.
         required: false
       ENV_VARS:
         description: Additional environment variables as a JSON formatted object.
+        required: false
+      CODECOV_TOKEN:
+        description: Codecov token.
         required: false
 
 jobs:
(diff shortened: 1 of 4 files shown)


.github/workflows/tests-unit-php.yml
```diff
@@ -68,3 +76,12 @@ jobs:
 
       - name: Run PHPUnit
         run: ./$(composer config bin-dir)/phpunit ${{ inputs.PHPUNIT_ARGS }}
+
+      - name: Upload coverage to Codecov
+        if: ${{ hashFiles(inputs.CODECOV_COVERAGE_FILE) != '' }}
```
dottxado: **question**

Given that I don't want to upload the coverage file, so I don't add the secret with the token, is it right that I must pass 
```bash
CODECOV_COVERAGE_FILE=''
```
otherwise the workflow will output an error because of the missing token during the upload step? Or are we planning to add an organization secret to "gently invite" a default behavior of uploading the coverage file?

tyrann0us: Good point. Now that I think about it, it’s probably better not to allow custom filenames, but to use `coverage.xml` instead, especially in light of more company standards. I will update accordingly …

dottxado (approved):

PR #251, opened 2026-06-01
Expose artifact name as workflow output

Biont:
# Expose artifact name as workflow output in `build-and-distribute`

Add an `artifact` output to the build-and-distribute workflow so calling workflows can reference the generated artifact name without reconstructing it manually.

A dedicated step captures the artifact name into both GITHUB_ENV and GITHUB_OUTPUT, making it available to the upload step and to any downstream jobs via `needs.<job>.outputs.artifact`.

- [x] Tests for the changes have been added (for bug fixes/features)

`build-plugin-archive` already exposes the generated archive name as a `workflow_call` output (`artifact`). This PR brings the same behaviour to `build-and-distribute`.

## Changes

- **`build-and-distribute.yml`**: Added a `Set artifact name` step that writes the artifact name (`PACKAGE_NAME-PACKAGE_VERSION`) to both `$GITHUB_ENV` and `$GITHUB_OUTPUT`, ensuring a single computed value is used everywhere. Wired that step output up through job-level and workflow-level `outputs` blocks so callers can reference it via `needs.<job>.outputs.artifact`.
- **`docs/build-and-distribute.md`**: Added a "Workflow output" subsection under "Artifact Output" documenting the new output and showing how a downstream job can consume it with `actions/download-artifact`.

### DIFF
--- a/.github/workflows/build-and-distribute.yml
+++ b/.github/workflows/build-and-distribute.yml
@@ -80,11 +80,17 @@ on:
       ENV_VARS:
         description: Additional environment variables as a JSON formatted object.
         required: false
+    outputs:
+      artifact:
+        description: The name of the generated release artifact
+        value: ${{ jobs.build-and-distribute.outputs.artifact }}
 
 jobs:
   build-and-distribute:
     timeout-minutes: 10
     runs-on: ubuntu-latest
+    outputs:
+      artifact: ${{ steps.set-artifact-name.outputs.artifact }}
     # Used to check whether there are changes to commit
     env:
       HAS_GIT_CHANGES: ''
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-distribute.yml
```diff
@@ -80,11 +80,17 @@ on:
       ENV_VARS:
         description: Additional environment variables as a JSON formatted object.
         required: false
+    outputs:
+      artifact:
+        description: The name of the generated release artifact
```
tyrann0us:
```suggestion
        description: The name of the generated release artifact.
```

tyrann0us (approved):
Code-wise LGTM. Thanks for working on that!

mishautkin:
@Biont have a question:
`needs.<job>.outputs.artifact` only works when the job that produced it is defined within the same workflow file.
In project level `e2e-tests.yml`, there is no build job, since B&D runs in a completely separate workflow, triggered independently. So it seems I can't just take the artifact 🤔

Biont:
@mishautkin 

Just to be clear: In your scenario scenario (e2e triggered completely separate from build) this PR does not add a new failure mode, right?

You previously had to come up with a workaround solution (download artifact via run ID) and this PR -rather unfortunately- does not help you with that.
Am I understanding this correctly?

---

There does not appear to be a way to access the workflow outputs vie REST or gh cli so this remains in workaround territory for now

PR #252, opened 2026-06-01
Reuse B&D artifact in test-playwright.yml

mishautkin:

Feature

Steps for retreiving B&D artifacts are missing.

If TESTED_ARTIFACT_SLUG is provided, the workflow will download the artifact built by B&D and place it as a ready-to-install .zip in TESTED_ARTIFACT_DIR.

- Add inputs
    - BUILD_WORKFLOW_FILENAME (build-and-distribute.yml by default)
    - TESTED_ARTIFACT_DIR (`tests/qa/resources/files` by default)
    - TESTED_ARTIFACT_SLUG
- Add steps
    - Resolve tested artifact run ID (uses the triggering run ID when invoked via workflow_run, otherwise finds the latest successful B&D run on the current branch via gh run list)
    - Download tested artifact (downloads the artifact matching <slug>-* from the resolved run)
    - Rename tested artifact (strips the version suffix to normalize .zip filename to TESTED_ARTIFACT_SLUG for further usage in tests)

No

### DIFF
--- a/.github/workflows/test-playwright.yml
+++ b/.github/workflows/test-playwright.yml
@@ -92,17 +92,12 @@ on:
         default: 'build-and-distribute.yml'
         required: false
         type: string
-      TESTED_ARTIFACT_DIR:
-        description: Directory to download the tested artifact into.
-        default: 'tests/qa/resources/files'
-        required: false
-        type: string
-      TESTED_ARTIFACT_SLUG:
+      ARTIFACT_NAME:
         description: >
-          Plugin slug used to download and repack the tested artifact as a zip for installation.
-          Expects the B&D artifact structure `<slug>-*/` containing `<slug>/` inside.
-          Produces `<slug>.zip` with `<slug>/` at the zip root and cleans up the downloaded directory.
-          Requires TESTED_ARTIFACT_DIR to also be set.
+          Name of the artifact produced by the build workflow to download and repack as a zip for plugin installation.
+          Expects the artifact to contain `<name>/` inside, produces `<name>.zip` with `<name>/` at the zip root.
+          When provided, the workflow resolves the producing run ID from the triggering workflow_run event or
+          the latest successful run of BUILD_WORKFLOW_FILENAME on the current branch.
         default: ''
         required: false
         type: string
@@ -285,9 +280,9 @@ jobs:
           npx wp-env run tests-cli wp config set WP_HOME "$NGROK_URL"
           sed -i "s|WP_BASE_URL=.*|WP_BASE_URL=$NGROK_URL|" .env.ci
 
-      - name: Resolve tested artifact run ID
+      - name: Resolve artifact run ID
         id: resolve-artifact-run-id
-        if: ${{ inputs.TESTED_ARTIFACT_SLUG != '' }}
+        if: ${{ inputs.ARTIFACT_NAME != '' }}
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/test-playwright.yml
```diff
@@ -87,6 +87,24 @@ on:
...
+        description: Workflow filename to resolve the build run ID from when triggered outside of a workflow_run event.
+        default: 'build-and-distribute.yml'
+        required: false
+        type: string
+      TESTED_ARTIFACT_DIR:
+        description: Directory to download the tested artifact into.
+        default: 'tests/qa/resources/files'
+        required: false
+        type: string
+      TESTED_ARTIFACT_SLUG:
```
Biont: Why the past-tense `TESTED_*`? Aren't we pointing at the _testee/subject-under-test_ here? It is not tested **yet**.

I also think this is not strictly a **slug** in that the name can be completely arbitrary. `upload-artifact` [simply refers to name here](https://github.com/actions/upload-artifact#inputs) so I think we should follow

.github/workflows/test-playwright.yml
```diff
@@ -87,6 +87,24 @@ on:
         default: ''
         required: false
         type: string
+      BUILD_WORKFLOW_FILENAME:
+        description: Workflow filename to resolve the build run ID from when triggered outside of a workflow_run event.
+        default: 'build-and-distribute.yml'
+        required: false
+        type: string
+      TESTED_ARTIFACT_DIR:
```
Biont: Is there a use-case for this to be configurable? It feels entirely internal to me.

.github/workflows/test-playwright.yml
```diff
@@ -87,6 +87,24 @@ on:
         default: ''
         required: false
         type: string
+      BUILD_WORKFLOW_FILENAME:
```
Biont, line 90: This one is odd in my opinion. It seems to be used as a fallback when no `TESTED_ARTIFACT_SLUG` is passed (because it is optional).

But the fallback hinges on the assumption that whatever **other** (non-build-and-distribute) build workflow **also** produces an artifact. **And** that said artifact **cannot** be referenced with a predictable name (convention or workflow output)

Is there even a second build workflow where these constraints apply?

Maybe I lack the full picture, but my gut feeling would be to make this _"named artifact only"_ and reduce the complexity.

.github/workflows/test-playwright.yml
```diff
@@ -267,6 +285,49 @@ jobs:
           npx wp-env run tests-cli wp config set WP_HOME "$NGROK_URL"
           sed -i "s|WP_BASE_URL=.*|WP_BASE_URL=$NGROK_URL|" .env.ci
 
+      - name: Resolve tested artifact run ID
+        id: resolve-artifact-run-id
+        if: ${{ inputs.TESTED_ARTIFACT_SLUG != '' }}
```
Biont: This is supposed to run when the slug is **NOT** passed, right?

Biont (changes requested):

.github/workflows/test-playwright.yml
mishautkin: In Mollie repo the WF which calls B&D is:

Name: Create release package
Filename: release.yml

.github/workflows/test-playwright.yml
mishautkin: Opposite 🤔: supposed to run only if the **slug is passed**. This WF is also used for tests of websites which are deployed to static test envs - in this case there's no artifact

mishautkin:
Hi, @Biont! As far as I understand this PR is not actual any more because of your updates in B&D (shortly: if the code didn't change and the artifact is present - B&D won't build a new one). So I suppose I can close this draft?

PR #253, opened 2026-06-03
Add duplicate run detection to skip redundant builds

Biont:
# Skip redundant builds in `build-and-distribute` when SHA is already compiled

Skip the entire compilation pipeline when the build branch already contains a build for the current commit SHA, saving several minutes of Node/PHP setup on duplicate workflow runs (re-runs, race conditions between concurrent triggers, or any other scenario where the same SHA is built more than once).

## Rationale

The intent of _build&distribute_ was always to maintain the build branch as the stable source of truth. It gives us a traceable, clean artifact that is not buried in a million GHA executions and it does not expire after 90 days.

**What we see in practice though** is that the coupling between `dev/foo`(<- the source branch) and `foo` (<- the build branch) is a surprisingly annoying obstacle to overcome when building reliable CI pipelines.
So there has always been an incentive to keep workflow artifacts around as a side-channel to the build-branch-commit (also because it offers a stable plugin/theme subfolder which we do not have on feature branches).

- On push to `dev/foo`, we cannot call e2e tests that require a built package since we push to the dev branch
- We cannot call e2e tests on `foo` because that has no direct connection to the dev branch: It will execute, but failures will not show up on the PR, so they'd be an costly, easy-to-miss optional signal that does not actually help QA efforts
- We do not want every build consumer to produce their own _build&distribute_ run because this is wasteful and redundant.
- However, we cannot properly decouple build & distribute from e2e tests and still chain them so that e2e only runs when we know that a build has completed. While it is possible to use a (completed) `workflow_run` trigger, the handover of the build branch checkout is still not trivial

A **clean solution** would be that indeed every consumer runs _build&distribute_ and then uses the artifact it produces. But as I said this is redundant and wasteful.
This PR attempts to make it less wasteful and thereby enable this as a clean pattern.

Ideally, it would turn a ~5min execution into a <1min execution that does not add any noise to the actual build branch, so it would be safe to call it repeatedly.

## What kind of change does this PR introduce?

Feature / optimization

## What is the current behavior?

Every workflow run performs a full build regardless of whether the build branch already contains compiled assets for the triggering commit. Two runs for the same SHA produce identical output but consume identical runner time.

### Duplicate run detection

A new "Check if already built for this SHA" step runs immediately before the build branch is checked out. It reads the remote build branch without checking it out and looks for the current commit SHA in:

- `SHA:` header in `style.css` — WordPress themes
- `SHA:` header in the main plugin PHP file — WordPress plugins

If the SHA is found, `SKIP_BUILD=true` is set and all compilation steps are skipped. The workflow then checks out the existing build branch (without wiping it) and proceeds directly to artifact packaging and upload, so the `artifact` output is always populated for downstream jobs.

Library projects (no theme or plugin header) are unaffected and always rebuild.

### SHA upsert

The `SHA:` field is now written into plugin/theme headers even when it is absent from the source file. Previously the `sed` substitution was a no-op if the field did not exist, which would prevent the skip gate from ever firing. The new behaviour inserts `SHA: <hash>` after the `Version:` line on the first build, enabling duplicate detection for all subsequent runs.

### Earlier preparation steps

"Determine package type" and "Prepare WordPress Plugin/Theme" have been moved to immediately after the source checkout, before any PHP or Node toolchain setup. This means a skip-gate hit now short-circuits the run before the most expensive steps (PHP setup, Composer install, Node setup, `npm ci`, `npm run build`) are even reached.

## Does this PR introduce a breaking change?

No. The `artifact` output and build branch push behaviour are unchanged for normal (non-duplicate) runs.

## Changes

- **`.github/workflows/build-and-distribute.yml`**:
  - Added "Check if already built for this SHA" step (after branch validation, before checkout)
  - Modified "Checkout the build branch with clean slate" to use a fast path when `SKIP_BUILD=true`: checks out the existing build branch without wiping it
  - Moved "Determine package type", "Prepare WordPress Plugin", "Prepare WordPress Theme" to immediately after checkout (before toolchain setup); all three now have `if: SKIP_BUILD != 'true'`
  - Updated SHA `sed` in both Prepare steps to upsert: replaces existing `SHA:` line or inserts after `Version:` if absent
  - Added `if: ${{ env.SKIP_BUILD != 'true' }}` to every build/compile step from "Check for composer.json" through "Git add, commit, and push"; artifact steps are left unconditional
- **`docs/build-and-distribute.md`**:
  - Updated intro step list to reflect the new skip check (step 2) and SHA upsert
  - Added "Duplicate run detection" subsection under "Build process details" explaining the gate, its scope, and the library limitation
  - Clarified `SHA:` behaviour in "Version Management" subsection

### DIFF
--- a/.github/workflows/build-and-distribute.yml
+++ b/.github/workflows/build-and-distribute.yml
@@ -146,42 +146,42 @@ jobs:
         run: |
           # Fetch latest public release
           LATEST_RELEASE=$(gh release list --limit 1 --json tagName --jq '.[0].tagName' || echo "")
-          
+
           if [ -z "$LATEST_RELEASE" ]; then
             echo "No releases found, using default version 0.0.0."
             LATEST_RELEASE="0.0.0"
           else
             echo "Latest release found: $LATEST_RELEASE"
           fi
-          
+
           # Get short SHA (first 7 characters)
           SHORT_SHA=$(echo "${{ github.sha }}" | cut -c1-7)
           echo "Short SHA: $SHORT_SHA"
-          
+
           # Strip 'dev/' prefix from branch name for version string (if it exists)
           ORIGIN_BRANCH="${{ github.ref_name }}"
           VERSION_BRANCH="${ORIGIN_BRANCH#dev/}"
-          
+
           # Normalize branch name for semver pre-release identifier
           # Replace invalid characters with hyphens and convert to lowercase
           NORMALIZED_BRANCH=$(echo "$VERSION_BRANCH" | sed 's/[^a-zA-Z0-9-]/-/g' | sed 's/--*/-/g' | sed 's/^-\|-$//g' | tr '[:upper:]' '[:lower:]')
-          
+
           # Construct semver with pre-release identifier using short SHA
           PACKAGE_VERSION="${LATEST_RELEASE}+${NORMALIZED_BRANCH}.${SHORT_SHA}"
           echo "Generated package version: $PACKAGE_VERSION"
-          
+
           echo "PACKAGE_VERSION=$PACKAGE_VERSION" >> $GITHUB_ENV
 
       - name: Inspect origin branch and determine build branch
         run: |
           ORIGIN_BRANCH="${{ github.ref_name }}"
           echo "Origin branch: $ORIGIN_BRANCH"
-          
+
           # Branch scheme:
           # dev/main -> main (stable production code with assets, where releases are created)
           # dev/ABC-123 -> ABC-123 (feature branch with compiled assets)
           # dev/xyz -> xyz (any other branch)
-          
+
           # Determine the build branch name
           if [ -n "${{ inputs.BUILT_BRANCH_NAME }}" ]; then
             BUILD_BRANCH="${{ inputs.BUILT_BRANCH_NAME }}"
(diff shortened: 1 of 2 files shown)


tyrann0us (approved):
>A new "Check if already built for this SHA" step runs immediately before the build branch is checked out. It reads the remote build branch without checking it out (`fetch-depth: 0` makes all remote objects available) and looks for the current commit SHA in […]:

I was about to ask you to update the PR description, since the current approach no longer uses `fetch-depth`, but then I checked your original commit [`bbdeb88` (this PR)](https://github.com/inpsyde/reusable-workflows/pull/253/changes/bbdeb884bfbf12d2ae6856171135aec884e4cdba), and it doesn't use it either. 😆
But in any case, the description is outdated anyway because my changes [`bbdeb88..9ea6215` (this PR)](https://github.com/inpsyde/reusable-workflows/pull/253/changes/bbdeb884bfbf12d2ae6856171135aec884e4cdba..9ea621586f6988edc33450bc34f1f418105a783d) moved and merged steps.

I'm approving, but it would be great if you could update the description to avoid confusion later. Thanks! 🙏🏽

.github/workflows/build-and-distribute.yml
```diff
@@ -215,34 +215,89 @@ jobs:
...
+          SHA="${{ github.sha }}"
+          BUILD_BRANCH="${{ env.BUILT_BRANCH_NAME }}"
+          PACKAGE_FILE="${{ env.PACKAGE_FILE }}"
+
+          if [ -z "$PACKAGE_FILE" ]; then
+            echo "No package file (library) — proceeding with build..."
+          elif ! git show-ref -q "refs/remotes/origin/$BUILD_BRANCH"; then
+            echo "Build branch does not exist yet — proceeding with build..."
+          elif git show "origin/$BUILD_BRANCH:$PACKAGE_FILE" 2>/dev/null | grep -qF "SHA: $SHA"; then
+            echo "Already built: SHA $SHA found in $PACKAGE_FILE on build branch."
```
tyrann0us: Let's add all data points we have to the message.
```suggestion
            echo "Already built: SHA $SHA found in $PACKAGE_FILE on build branch $BUILD_BRANCH."
```
Sorry, only noticed this improvement now.

PR #256, opened 2026-07-20
Fix release lookup auth and make build-skip version-aware

zhyian:

---

Bug fix.

---

The "Determine package version" step sets `GITHUB_TOKEN` to an SSH private key (`secrets.GITHUB_USER_SSH_KEY`) and uses it for `gh release list`. An SSH key is a multi-line PEM blob, which can't be used as an HTTP Authorization header, so `gh` fails.

Separately, the "Check if already built for this SHA" step keys the build-skip cache on the commit SHA only. If a SHA was already built (e.g. with the wrong 0.0.0 version), a re-run, including one with a corrected or manually specified PACKAGE_VERSION, is skipped and the stale artifact is re-served, so the correct version never gets applied.

---

1 .The version-lookup step now authenticates `gh` with a real API token instead of the SSH key: an optional `GITHUB_USER_TOKEN` secret, falling back to the automatic `GITHUB_TOKEN` when not provided.
2. The "Check if already built for this SHA" step now compares both the committed SHA and the version. The build is only skipped when both match, a changed or manually specified `PACKAGE_VERSION` triggers a rebuild so the correct version is applied. This also self-fixes any artifact previously committed with the wrong version.

---

No

---

## Security impact

**What changed (security perspective):**

**Testing done:**

**Residual risk (link Jira if any):**

---

### DIFF
--- a/.github/workflows/build-and-distribute.yml
+++ b/.github/workflows/build-and-distribute.yml
@@ -78,7 +78,7 @@ on:
         description: Public SSH key associated with the GitHub user passed as `GITHUB_USER_NAME`.
         required: true
       GITHUB_USER_TOKEN:
-        description: Authentication token with write permission needed by the release bot (falls back to `GITHUB_TOKEN`).
+        description: Authentication token with read permission needed by the release bot (falls back to `GITHUB_TOKEN`).
         required: false
       ENV_VARS:
         description: Additional environment variables as a JSON formatted object.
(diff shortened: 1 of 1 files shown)


cristianobaptista (approved):
LGTM after testing internally in a couple of packages that were having this specific issue.

Package version is now being set correctly based on latest release and falling back to `0.0.0` as it should if no releases exist. The package is also not rebuilt as described.

PR #257, opened 2026-07-24
Add OpenVPN composite action

dottxado:
- [] Tests for the changes have been added (for bug fixes/features)

---

Feature

---

- The deploy workflow uses WireGuard

---

- Expose a composite action to establish VPN connections using OpenVPN protocol
- Migrate `deploy-deployer.yml` to use the new composite action

---

Yes, for the `deploy-deployer` usage, current workflows that use WireGuard will need to transition to the new architecture first.

---

## Security impact

None

---

### DIFF
--- a/.github/actions/setup-openvpn/README.md
+++ b/.github/actions/setup-openvpn/README.md
@@ -14,7 +14,7 @@ steps:
       OVPN_CONFIG: ${{ secrets.OVPN_CONFIG }}
       OVPN_USERNAME: ${{ secrets.OVPN_USERNAME }}
       OVPN_PASSWORD: ${{ secrets.OVPN_PASSWORD }}
-      OVPN_GATEWAY_IP: ${{ secrets.OVPN_GATEWAY_IP }}
+      OVPN_GATEWAY_IP: ${{ vars.OVPN_GATEWAY_IP }}
     if: ${{ env.OVPN_CONFIG != '' && env.OVPN_USERNAME != '' && env.OVPN_PASSWORD != '' && env.OVPN_GATEWAY_IP != '' }}
     with:
       ovpn-config: ${{ env.OVPN_CONFIG }}
@@ -45,7 +45,7 @@ steps:
       OVPN_CONFIG: ${{ secrets.OVPN_CONFIG }}
       OVPN_USERNAME: ${{ secrets.OVPN_USERNAME }}
       OVPN_PASSWORD: ${{ secrets.OVPN_PASSWORD }}
-      OVPN_GATEWAY_IP: ${{ secrets.OVPN_GATEWAY_IP }}
+      OVPN_GATEWAY_IP: ${{ vars.OVPN_GATEWAY_IP }}
     if: ${{ env.OVPN_CONFIG != '' && env.OVPN_USERNAME != '' && env.OVPN_PASSWORD != '' && env.OVPN_GATEWAY_IP != '' }}
     with:
       ovpn-config: ${{ env.OVPN_CONFIG }}
--- a/.github/actions/setup-openvpn/README.md
+++ b/.github/actions/setup-openvpn/README.md
@@ -1,0 +1,46 @@
+# Set up OpenVPN
+
+This composite action installs OpenVPN and establishes a VPN connection using the provided configuration and credentials. It should be used before any step that sets up SSH keys for remote hosts behind a VPN.
+
+## Simple usage example
+
+```yml
+steps:
+  - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+  - uses: ./.github/actions/setup-openvpn
+    with:
+      ovpn-config: ${{ secrets.OVPN_CONFIG }}
+      ovpn-username: ${{ secrets.OVPN_USERNAME }}
+      ovpn-password: ${{ secrets.OVPN_PASSWORD }}
+```
+
+## Inputs
+
+| Name            | Default | Description                            |
+|-----------------|---------|----------------------------------------|
+| `ovpn-config`   | -       | OpenVPN configuration file content.    |
+| `ovpn-username` | -       | OpenVPN username for authentication.   |
+| `ovpn-password` | -       | OpenVPN password for authentication.   |
+
+## Usage with SSH
+
+The VPN connection must be established before configuring SSH keys for remote hosts.
+
+```yml
+steps:
+  - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+  - name: Set up OpenVPN
+    uses: inpsyde/reusable-workflows/.github/actions/setup-openvpn@main
+    with:
+      ovpn-config: ${{ secrets.OVPN_CONFIG }}
+      ovpn-username: ${{ secrets.OVPN_USERNAME }}
+      ovpn-password: ${{ secrets.OVPN_PASSWORD }}
+
+  - name: Configure SSH key
+    run: |
+      mkdir -p ~/.ssh
+      ssh-keyscan -p ${{ secrets.DEPLOY_PORT }} ${{ secrets.DEPLOY_HOSTNAME }} >> ~/.ssh/known_hosts
+
+  # Deployment steps can now reach hosts behind the VPN
+```
--- a/.github/actions/setup-openvpn/README.md
+++ b/.github/actions/setup-openvpn/README.md
@@ -1,6 +1,8 @@
 # Set up OpenVPN
 
-This composite action installs OpenVPN and establishes a VPN connection using the provided configuration and credentials. It should be used before any step that sets up SSH keys for remote hosts behind a VPN.
+This composite action installs OpenVPN and establishes a VPN connection using the provided
+configuration and credentials. It should be used before any step that sets up SSH keys for remote
+hosts behind a VPN.
 
 ## Simple usage example
 
@@ -8,19 +10,27 @@ This composite action installs OpenVPN and establishes a VPN connection using th
 steps:
   - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
   - uses: ./.github/actions/setup-openvpn
+    env:
+      OVPN_CONFIG: ${{ secrets.OVPN_CONFIG }}
+      OVPN_USERNAME: ${{ secrets.OVPN_USERNAME }}
+      OVPN_PASSWORD: ${{ secrets.OVPN_PASSWORD }}
+      OVPN_GATEWAY_IP: ${{ secrets.OVPN_GATEWAY_IP }}
+    if: ${{ env.OVPN_CONFIG != '' && env.OVPN_USERNAME != '' && env.OVPN_PASSWORD != '' && env.OVPN_GATEWAY_IP != '' }}
     with:
-      ovpn-config: ${{ secrets.OVPN_CONFIG }}
-      ovpn-username: ${{ secrets.OVPN_USERNAME }}
-      ovpn-password: ${{ secrets.OVPN_PASSWORD }}
+      ovpn-config: ${{ env.OVPN_CONFIG }}
+      ovpn-username: ${{ env.OVPN_USERNAME }}
+      ovpn-password: ${{ env.OVPN_PASSWORD }}
+      ovpn-gateway-ip: ${{ env.OVPN_GATEWAY_IP }}
 ```
 
 ## Inputs
 
-| Name            | Default | Description                            |
-|-----------------|---------|----------------------------------------|
-| `ovpn-config`   | -       | OpenVPN configuration file content.    |
-| `ovpn-username` | -       | OpenVPN username for authentication.   |
-| `ovpn-password` | -       | OpenVPN password for authentication.   |
+| Name              | Default | Description                             |
+|-------------------|---------|-----------------------------------------|
+| `ovpn-config`     | `''`    | OpenVPN configuration file content.     |
+| `ovpn-username`   | `''`    | OpenVPN username for authentication.    |
+| `ovpn-password`   | `''`    | OpenVPN password for authentication.    |
+| `ovpn-gateway-ip` | `''`    | Expected IP address of the VPN gateway. |
 
 ## Usage with SSH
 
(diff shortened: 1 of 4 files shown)


.github/actions/setup-openvpn/README.md
```diff
@@ -0,0 +1,46 @@
+# Set up OpenVPN
```
severinil, line 1: Since we tested this with the real org secrets (`NORDLAYER_OVPN_TCP_CONFIG`, `NORDLAYER_USERNAME`, `NORDLAYER_PASSWORD` → `OVPN_CONFIG`, `OVPN_USERNAME`, `OVPN_PASSWORD`), can we add that mapping to the README? Right now the example only shows generic `secrets.OVPN_CONFIG` with no link to the actual org secret names, so anyone setting this up in another repo has to rediscover the mapping.

One open question: should we keep the generic `OVPN_*` names (keeps this reusable if we ever switch VPN providers again) and just document the mapping, or use the `NORDLAYER_*` names directly in the action for simplicity? Leaning toward keeping generic + documenting, but curious if you see it differently.

.github/actions/setup-openvpn/action.yml
```diff
@@ -0,0 +1,35 @@
...
+  using: composite
+  steps:
+    - name: Install OpenVPN
+      run: |
+        sudo apt update
+        sudo apt install -y openvpn openvpn-systemd-resolved
+      shell: bash
+
+    - name: Write OpenVPN config
+      run: |
```
severinil, line 30: Per GitHub's hardening guide, secret expressions should go through an intermediate env var instead, to avoid shell injection if the value ever contains special characters. Reference: https://docs.github.com/en/enterprise-server@3.10/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable

```
env:
  OVPN_CONFIG: ${{ inputs.ovpn-config }}
run: |
  echo "$OVPN_CONFIG" > /tmp/client.ovpn
  chmod 600 /tmp/client.ovpn
```

.github/workflows/deploy-deployer.yml
```diff
@@ -77,13 +83,15 @@ jobs:
         with:
           ssh-private-key: ${{ secrets.GITHUB_USER_SSH_KEY }}
 
-      - name: Set up WireGuard
-        uses: inpsyde/actions/setup-wireguard@v1
+      - name: Set up OpenVPN
+        uses: inpsyde/reusable-workflows/.github/actions/setup-openvpn@main
         env:
-          WIREGUARD_CONFIGURATION: ${{ secrets.WIREGUARD_CONFIGURATION }}
```
severinil, line 83: This removes `WIREGUARD_CONFIGURATION` entirely. If any other repo currently calls `deploy-deployer.yml` with that secret, their workflow will fail on the next run. We must wait until the VPN transition is over before merging this PR.

.github/actions/setup-openvpn/action.yml
```diff
@@ -0,0 +1,35 @@
...
+        echo "${{ inputs.ovpn-config }}" > /tmp/client.ovpn
+        chmod 600 /tmp/client.ovpn
+      shell: bash
+
+    - name: Connect
+      uses: "kota65535/github-openvpn-connect-action@cd2ed8a90cc7b060dc4e001143e811b5f7ea0af5" # v3.1.0
+      with:
+        config_file: /tmp/client.ovpn
+        username: ${{ inputs.ovpn-username }}
+        password: ${{ inputs.ovpn-password }}
```
severinil, line 40: Can we add a step after `Connect` to verify the egress IP matches the expected gateway IP? Without it, a dropped or misconfigured VPN connection could let deploy traffic go out over the runner's normal IP instead of failing loudly — which defeats the point of the IP-allowlist setup this is built for.

I have already stored the IP as an org-level GitHub Actions variable: `NORDLAYER_BOTS_GATEWAY_IP`.

```
- name: Verify VPN egress IP
  run: |
    IP=$(curl -s https://api.ipify.org)
    echo "Egress IP: $IP"
    [[ "$IP" == "${{ vars.NORDLAYER_BOTS_GATEWAY_IP }}" ]] || { echo "Unexpected egress IP: $IP"; exit 1; }
  shell: bash
```

Sorry if I originally told you this step was optional. Just figured out this is actually needed — without it, the VPN client could report "connected" while traffic still silently exits over the runner's normal IP, and the job would have no way of knowing. The workflow needs to fail in that case, not deploy anyway.

.github/workflows/deploy-deployer.yml
```diff
@@ -77,13 +83,15 @@ jobs:
...
-        if: ${{ env.WIREGUARD_CONFIGURATION != '' }}
+          OVPN_CONFIG: ${{ secrets.OVPN_CONFIG }}
+        if: ${{ env.OVPN_CONFIG != '' }}
```
severinil: Minor: the `if:` condition only checks `OVPN_CONFIG` is non-empty. If `OVPN_CONFIG` is set but username/password aren't, this will fail with a confusing auth error (deep in the connect action) instead of a clear message.

severinil (changes requested):

.github/actions/setup-openvpn/README.md
dottxado: Since this repository is public, we shouldn't reference internal secrets in the docs. That being said, I believe this is still an issue for our teammates, as the connection with our organization is not always obvious. If we take any action to solve this, we should take a deliberate approach to all workflows, not just this.

.github/workflows/deploy-deployer.yml
dottxado: Yes, I will point this out in the PR description :muscle:

.github/actions/setup-openvpn/action.yml
dottxado: No worries, added :muscle:

.github/actions/setup-openvpn/action.yml
dottxado: Done :muscle:

dottxado: Please check if I have used it correctly: I initially put `echo "$OVPN_CONFIG"` like your suggestion, but then I decided to align the variable call as the others.

.github/workflows/deploy-deployer.yml
dottxado: Right :woman_facepalming: Added all the vars, including the new gateway IP one.

.github/actions/setup-openvpn/action.yml
severinil: Thanks @dottxado!

.github/actions/setup-openvpn/action.yml
severinil: Cool, thanks @dottxado! :)

.github/actions/setup-openvpn/action.yml
severinil: If I got the official doc right, we must reference the shell variable, not the expression:
```
- name: Write OpenVPN config
  env:
    OVPN_CONFIG: ${{ inputs.ovpn-config }}
  run: |
    echo "$OVPN_CONFIG" > /tmp/client.ovpn
    chmod 600 /tmp/client.ovpn
  shell: bash
```

.github/actions/setup-openvpn/action.yml
```diff
@@ -0,0 +1,52 @@
...
+        OVPN_PASSWORD: ${{ inputs.ovpn-password }}
+      with:
+        config_file: /tmp/client.ovpn
+        username: ${{ env.OVPN_USERNAME }}
+        password: ${{ env.OVPN_PASSWORD }}
+
+    - name: Verify VPN egress IP
+      env:
+        GATEWAY_IP: ${{ inputs.ovpn-gateway-ip }}
+      run: |
```
severinil, line 45: If I got the official doc right, we must reference the shell variable here as well, not the expression:
```
- name: Verify VPN egress IP
  env:
    GATEWAY_IP: ${{ inputs.ovpn-gateway-ip }}
  run: |
    IP=$(curl -s https://api.ipify.org)
    echo "Egress IP: $IP"
    [[ "$IP" == "$GATEWAY_IP" ]] || { echo "Unexpected egress IP: $IP"; exit 1; }
  shell: bash
```

.github/actions/setup-openvpn/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+```yml
+steps:
+  - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+  - uses: ./.github/actions/setup-openvpn
+    env:
+      OVPN_CONFIG: ${{ secrets.OVPN_CONFIG }}
+      OVPN_USERNAME: ${{ secrets.OVPN_USERNAME }}
+      OVPN_PASSWORD: ${{ secrets.OVPN_PASSWORD }}
+      OVPN_GATEWAY_IP: ${{ secrets.OVPN_GATEWAY_IP }}
```
severinil: Gateway IP is `secrets`, should be `vars`.

docs/deploy-deployer.md
```diff
@@ -86,8 +86,11 @@ jobs:
 | `DEPLOY_USER`             | Yes      | SSH user on the target server                                                            |
 | `GITHUB_USER_SSH_KEY`     | Yes      | Private SSH key used for repository checkout and remote server access                    |
 | `COMPOSER_AUTH_JSON`      | Yes      | Authentication for privately hosted packages and repositories as a JSON formatted object |
-| `WIREGUARD_CONFIGURATION` | No       | The full content of the WireGuard configuration file for VPN tunnel setup                |
 | `NPM_REGISTRY_TOKEN`      | No       | Authentication for the private npm registry                                              |
+| `OVPN_CONFIG`             | No       | The full content of the OpenVPN configuration file                                       |
+| `OVPN_USERNAME`           | No       | The username to authenticate with the OpenVPN server                                     |
+| `OVPN_PASSWORD`           | No       | The password to authenticate with the OpenVPN server                                     |
+| `OVPN_GATEWAY_IP`         | No       | Authentication for the private npm registry                                              |
```
severinil: Description doesn't match the field :)

.github/actions/setup-openvpn/README.md
severinil: Good point, makes sense to keep `NORDLAYER_*` out of here.
I will document the mapping in the VPN section of the Eng Handbook instead.

severinil:
@dottxado Nice to have, if not too complicated: we should delete the OpenVPN configuration file from the runner once the VPN connection is not needed anymore.
```
      - name: Delete OpenVPN config
        if: always()
        run: rm -f /tmp/client.ovpn
```

docs/deploy-deployer.md
dottxado: Opsie :grin: 
Fixed [here](19a8fb79c90cc8b27afb49787fd19e219e396e64), but then moved to vars [here](4934ead58426484da1afbe07fb7ff8850bf46423)

.github/actions/setup-openvpn/README.md
dottxado: Ah sorry, I expected it to be a secret, and I didn't check where you added it. Fixed [here](https://github.com/inpsyde/reusable-workflows/pull/257/changes/4934ead58426484da1afbe07fb7ff8850bf46423)

.github/actions/setup-openvpn/action.yml
dottxado: Right :muscle: 
Fixed [here](https://github.com/inpsyde/reusable-workflows/pull/257/changes/efd14d86974bcb6d6f325048a99ca1799e9b3622): I also did a small cleanup.

.github/actions/setup-openvpn/action.yml
```diff
@@ -47,3 +47,8 @@ runs:
         echo "Egress IP: $IP"
         [[ "$IP" == "$GATEWAY_IP" ]] || { echo "Unexpected egress IP: $IP"; exit 1; }
       shell: bash
+
+    - name: Delete OpenVPN config
+      if: always()
+      run: rm -f /tmp/client.ovpn
+      shell: bash
```
tyrann0us, line 54: Out of curiosity, is there any difference to the `trap` approach used in the legacy WireGuard action?
https://github.com/inpsyde/actions/blob/main/setup-wireguard/action.yml#L16-L17

I don't have a preference for any approach. 🙂

dottxado: In this case yes, but only because we have separate steps for file creation and connecting to the VPN. `trap` will execute the command at the closing of the shell, and since shells are closed for each step, adding `trap` in the same step where we create the file will also delete it before it is used by the `Connect` step :smile:

severinil (approved):

.github/workflows/deploy-deployer.yml
```diff
@@ -77,13 +87,19 @@ jobs:
         with:
           ssh-private-key: ${{ secrets.GITHUB_USER_SSH_KEY }}
 
-      - name: Set up WireGuard
```
severinil, line 80: For a safe transition, I suggest we keep both steps side by side for now, so existing consumers keep working while others migrate.

```
- name: Set up WireGuard
  uses: inpsyde/actions/setup-wireguard@v1
  env:
    WIREGUARD_CONFIGURATION: ${{ secrets.WIREGUARD_CONFIGURATION }}
  if: ${{ env.WIREGUARD_CONFIGURATION != '' }}

- name: Set up OpenVPN
  uses: inpsyde/reusable-workflows/.github/actions/setup-openvpn@main
  env:
    OVPN_CONFIG: ${{ secrets.OVPN_CONFIG }}
    OVPN_USERNAME: ${{ secrets.OVPN_USERNAME }}
    OVPN_PASSWORD: ${{ secrets.OVPN_PASSWORD }}
    OVPN_GATEWAY_IP: ${{ inputs.OVPN_GATEWAY_IP }}
  if: ${{ env.WIREGUARD_CONFIGURATION == '' && env.OVPN_CONFIG != '' && env.OVPN_USERNAME != '' && env.OVPN_PASSWORD != '' && env.OVPN_GATEWAY_IP != '' }}
```
The `WIREGUARD_CONFIGURATION == ''` check on the OpenVPN step makes sure only one VPN client ever runs, even if a repo accidentally ends up with both sets of secrets configured during migration.

For the same reason, please keep `WIREGUARD_CONFIGURATION` in the `secrets:` declaration block and its row in `docs/deploy-deployer.md` (we can mark it deprecated) rather than removing it, so existing consumers don't break.

Once everyone has migrated to NordLayer, we can remove WireGuard support in the follow-up ticket SDO-1156.

severinil (changes requested):

.github/workflows/deploy-deployer.yml
dottxado: Sounds legit :muscle: Done [here](https://github.com/inpsyde/reusable-workflows/pull/257/commits/3e67e9ecc3dc90d23f0100803ecd0030957d683e)

.github/workflows/deploy-deployer.yml
```diff
@@ -82,8 +95,21 @@ jobs:
...
+        uses: inpsyde/reusable-workflows/.github/actions/setup-openvpn@main
+        env:
+          OVPN_CONFIG: ${{ secrets.OVPN_CONFIG }}
+          OVPN_USERNAME: ${{ secrets.OVPN_USERNAME }}
+          OVPN_PASSWORD: ${{ secrets.OVPN_PASSWORD }}
+          OVPN_GATEWAY_IP: ${{ inputs.OVPN_GATEWAY_IP }}
+          WIREGUARD_CONFIGURATION: ${{ secrets.WIREGUARD_CONFIGURATION }}
+        if: ${{ env.WIREGUARD_CONFIGURATION == '' && env.OVPN_CONFIG != '' && env.OVPN_USERNAME != '' && env.OVPN_PASSWORD != '' && env.OVPN_GATEWAY_IP != '' }}
         with:
-          wireguard-configuration: ${{ secrets.WIREGUARD_CONFIGURATION }}
```
tyrann0us, line 86: Careful, with the back and forth, `WIREGUARD_CONFIGURATION` now wrongly is an env and not a secret anymore.

dottxado: Since it is used as a condition in an `if`, it needs first to be defined as an `env` variable. GitHub will mask it even if it is an env variable, because it comes from a `secret` :muscle:

tyrann0us: I should shut up … 🙈  I only saw that the line has moved and didn't check the rest at all. Thought it might have been a mistake during reverting. Sorry for the noise!

severinil (approved):

.github/workflows/deploy-deployer.yml
severinil: Thanks @dottxado!

.github/workflows/deploy-deployer.yml
severinil, line 86: The "Set up WireGuard" step lost its `with:` block — comparing against `main`, it used to have:
```
with:
  wireguard-configuration: ${{ secrets.WIREGUARD_CONFIGURATION }}
```
Can we add it back?

severinil (changes requested):

severinil (approved):

.github/workflows/deploy-deployer.yml
dottxado: Done [here](https://github.com/inpsyde/reusable-workflows/pull/257/commits/3f763b608e887da859e3429fb9649d8933f80cce)

PR #262, opened 2026-08-11
Fix: verify VPN egress IP via a tunnel-routed probe

amiut:

* [x] The commit message follows our guidelines
* [ ] Tests for the changes have been added (for bug fixes/features)
* [ ] Docs have been added/updated (for bug fixes/features)

---

Bug fix

---

**Workflows using `setup-openvpn` fail during the `Verify VPN egress IP` step.**

The verification step currently determines the workflow runner's public IP by calling a third-party HTTPS IP-echo service and comparing the returned address with the expected VPN gateway IP.

This works only when the VPN routes **all outbound traffic** through the tunnel.

Some VPN profiles use split tunneling and route only specific destinations through the VPN while leaving the runner's default route unchanged. In that case, requests to arbitrary internet hosts — including the IP-echo service — bypass the VPN and use the runner's normal network interface.

As a result, the verification step reports the runner's public IP instead of the VPN gateway IP and fails, even though the VPN tunnel itself has been established successfully and the configured destinations are correctly routed through it.

One possible workaround would be to add the IP-echo service to the VPN destinations list. However, this is unreliable for services hosted behind Cloudflare because their IP addresses may change and cannot be safely hard-coded.

---

Replaced the IP detection logic with a simpler approach.

The verification step now determines the public IP using Cloudflare's DNS-based IP echo service:

```bash
IP=$(dig +short @1.1.1.1 whoami.cloudflare CH TXT | tr -d '"')
```

This allows the egress IP check to work without depending on the dynamically resolved IP addresses of an external HTTPS service.

**Note:** `1.1.1.1` must still be included in the VPN destinations list so that this DNS request is routed through the tunnel.

`dnsutils` is already installed on the runner image, so no additional installation step is required.

No inputs or existing interfaces were changed, so callers are unaffected.

---

No.

---

## Security impact

None.

---

### DIFF
--- a/.github/actions/setup-openvpn/action.yml
+++ b/.github/actions/setup-openvpn/action.yml
@@ -43,7 +43,7 @@ runs:
       env:
         GATEWAY_IP: ${{ inputs.ovpn-gateway-ip }}
       run: |
-        IP=$(curl -s https://api.ipify.org)
+        IP=$(dig +short @1.1.1.1 whoami.cloudflare ch txt | tr -d '"')
         echo "Egress IP: $IP"
         [[ "$IP" == "$GATEWAY_IP" ]] || { echo "Unexpected egress IP: $IP"; exit 1; }
       shell: bash
--- a/.github/workflows/deploy-deployer.yml
+++ b/.github/workflows/deploy-deployer.yml
@@ -99,7 +99,7 @@ jobs:
           wireguard-configuration: ${{ secrets.WIREGUARD_CONFIGURATION }}
 
       - name: Set up OpenVPN
-        uses: inpsyde/reusable-workflows/.github/actions/setup-openvpn@fix/setup-openvpn-verify-egress-ip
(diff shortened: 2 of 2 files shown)


Chrico (approved):

severinil:
Quick note for anyone wondering why 1.1.1.1 doesn't need a manual entry in our gateways' split-tunnel include lists (it isn't there today, on either gateway):

Checked a real PUSH_REPLY from a Bots-gateway connection — NordLayer auto-pushes host routes for 1.1.1.1 and 1.0.0.1 on every split-tunnel profile, regardless of what's configured in the include list. That's the platform keeping its own DNS resolvers reachable through the tunnel by default, not something we need to add ourselves.

Confirmed this is actually happening in production: a Bots-gateway run from 2026-08-19 shows the check resolving and matching correctly (`Egress IP: 94.101.112.25`).

PR #265, opened 2026-08-12
Fix: wordpress.org publishing workflow fails on rsync warning

### LINKED ISSUE #264: [Bug]: wordpress.org publishing workflow fails on rsync warning
### Description of the bug

Running workflow causes `rsync` warnings like these:
```
file has vanished: "/home/runner/work/paypal-point-of-sale/paypal-point-of-sale/svn_repository/zettle-pos-integration/trunk/vendor/container-interop/service-provider/puli.json"
file has vanished: "/home/runner/work/paypal-point-of-sale/paypal-point-of-sale/svn_repository/zettle-pos-integration/trunk/vendor/container-interop/service-provider/src/ServiceProviderInterface.php"
```
We are strict about any kind of errors in this workflow, so it stops with failure after this.

The problem seems to be that `rsync` destination is inside its source.

### Reproduction instructions

1. Go to [this action run](https://github.com/inpsyde/paypal-point-of-sale/actions/runs/31496859534/job/94019695978) (the first usage of the workflow)
2. Re-run the action
3. Observe errors

### Expected behavior

1. Action finishes without errors
2. All plugin files are pushed to wordpress.org SVN

### Environment info

_No response_

### Relevant log output

```shell
file has vanished: "/home/runner/work/paypal-point-of-sale/paypal-point-of-sale/svn_repository/zettle-pos-integration/trunk/vendor/dhii/collections-interface/src/ClearableContainerInterface.php"
file has vanished: "/home/runner/work/paypal-point-of-sale/paypal-point-of-sale/svn_repository/zettle-pos-integration/trunk/vendor/dhii/collections-interface/src/ContainerFactoryInterface.php"
...

Error: Process completed with exit code 24.
```

### Additional context

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

strangerkir:
### Please check if the PR fulfills these requirements

---

### What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)
Bug fix (and a bit of refactoring).

---

### What is the current behavior? (You can also link to an open issue here)
The GitHub Action using this workflow fails because of `rsync` warnings:
```shell
file has vanished: "/home/runner/work/paypal-point-of-sale/paypal-point-of-sale/svn_repository/zettle-pos-integration/trunk/vendor/container-interop/service-provider/puli.json"
file has vanished: "/home/runner/work/paypal-point-of-sale/paypal-point-of-sale/svn_repository/zettle-pos-integration/trunk/vendor/container-interop/service-provider/src/ServiceProviderInterface.php"
```

---

### What is the new behavior (if this is a feature change)?
No warnings from `rsync`.

Additionally, the `VALIDATED_SVN_PLUGIN_SLUG` variable is introduced instead of reading raw input every time. This adds clarity and consistency with the `VALIDATED_PLUGIN_VERSION` var we are using already.

---

### Does this PR introduce a breaking change? (What changes might users need to make in their application due to this PR?)
No.

---

### Security impact
None.

---

### Other information
Closes #264.

### DIFF
--- a/.github/workflows/wordpress-org-release.yml
+++ b/.github/workflows/wordpress-org-release.yml
@@ -63,7 +63,6 @@ jobs:
             exit 1
           fi
 
-          echo "VALIDATED_SVN_PLUGIN_SLUG=$SVN_PLUGIN_SLUG" >> "$GITHUB_ENV"
           echo "✅ Valid plugin slug: $SVN_PLUGIN_SLUG"
 
       - name: Validate versions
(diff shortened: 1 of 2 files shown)


Chrico (approved):

.github/workflows/wordpress-org-release.yml
```diff
@@ -63,6 +63,7 @@ jobs:
             exit 1
           fi
 
+          echo "VALIDATED_SVN_PLUGIN_SLUG=$SVN_PLUGIN_SLUG" >> "$GITHUB_ENV"
```
Chrico: **note:**

We should not write everything into the environment, since it makes it accessible for every action we consume. Why is it necessary to move this out of `inputs` and into the env?

strangerkir: This is not strictly necessary. I added this because it gives a clear idea that the input was validated already, and because it is consistent with the already existing `$VALIDATED_PLUGIN_VERSION` variable. Spotted this at the last moment as an inconsistency.

Do you think I should revert this?

UPD: I think I'll revert it here anyway, as it doesn't belong here. We can continue the discussion, but I'd rather create another PR if needed.

Chrico: It would be good to have only what is really necessary in' env'. Validation can still happen; you can `exit` with an error when the validation fails, which will stop the whole workflow. 👍🏻

strangerkir: I created a follow-up PR to remove the `$VALIDATED_PLUGIN_VERSION` variable
https://github.com/inpsyde/reusable-workflows/pull/274

PR #269, opened 2026-08-13
Fix SVN status missing changes

strangerkir:

---

Bug fix

---

Two checks piped svn status into grep -q. Since grep -q exits as soon as it finds a match, it can close the pipe while svn status (a separate, still-running process) is mid-write, killing it with SIGPIPE. Under set -o pipefail, that non-zero exit (not grep's own success) becomes the pipeline's result, so the `if`/`||` sees 'no match' even when grep actually found one. This only surfaces when there's something to actually match, which is why it went unnoticed.

In practice this means:
  - In 'Update SVN tracking', the deletion block is silently skipped, so files that should be removed from SVN trunk are never deleted.
 - In 'Commit changes to SVN repository', the 'anything to commit' guard can falsely report 'No changes in trunk' and abort the whole job, even when there are real pending changes.

---

`-q` option is removed from the grep call, so svn status always finishes writing before grep exits, and the pipeline's exit code correctly reflects whether grep found a match.

This may take a few seconds longer, but correctness is much more important in this case.

---

No.

---

## Security impact

None.

---

### DIFF
--- a/.github/workflows/wordpress-org-release.yml
+++ b/.github/workflows/wordpress-org-release.yml
@@ -234,7 +234,9 @@ jobs:
           [ -d "$SVN_REPO_ROOT" ] || { echo "❌ SVN_REPO_ROOT not found"; exit 1; }
           cd "$SVN_REPO_ROOT"
 
-          # No -q here either, same reason as in "Update SVN tracking".
+          # Do not add -q to grep: it exits on the first match, SIGPIPE-ing svn status mid-write;
+          # under pipefail that non-zero exit wins over grep's, which would wrongly report
+          # "No changes in trunk" and abort even when there are real changes to commit.
           svn status trunk | grep '^[ADMR!~]' >/dev/null || { echo "❌ No changes in trunk"; exit 1; }
 
           # WordPress.org SVN doesn't support SSH keys or tokens.
--- a/.github/workflows/wordpress-org-release.yml
+++ b/.github/workflows/wordpress-org-release.yml
@@ -209,7 +209,9 @@ jobs:
           svn add . --force
           
           #Only run if there's anything to delete
-          if svn status | grep -q '^!'; then
+          # No -q: it exits on the first match, SIGPIPE-ing svn status mid-write; under
+          # pipefail that non-zero exit wins over grep's, silently hiding real matches.
+          if svn status | grep '^!' >/dev/null; then
             svn status | grep '^!' | cut -c9- | while IFS= read -r file; do
               svn delete "$file@"  # @ suffix handles files with @ in name
             done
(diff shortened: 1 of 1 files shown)


Chrico (approved):

.github/workflows/wordpress-org-release.yml
```diff
@@ -209,7 +209,9 @@ jobs:
           svn add . --force
           
           #Only run if there's anything to delete
```
tyrann0us: **suggestion**

```suggestion
          # Only run if there's anything to delete.
```

.github/workflows/wordpress-org-release.yml
```diff
@@ -209,7 +209,9 @@ jobs:
...
-          if svn status | grep -q '^!'; then
+          # No -q: it exits on the first match, SIGPIPE-ing svn status mid-write; under
+          # pipefail that non-zero exit wins over grep's, silently hiding real matches.
```
tyrann0us: **suggestion**

Claude Code is very eager to comment as much as possible. This doesn't add value; no one will know later why the comment mentions `-q`. I recommend removing it; use the Git history if you are interested in the why. 
```suggestion
```

.github/workflows/wordpress-org-release.yml
```diff
@@ -233,7 +235,8 @@ jobs:
           [ -d "$SVN_REPO_ROOT" ] || { echo "❌ SVN_REPO_ROOT not found"; exit 1; }
           cd "$SVN_REPO_ROOT"
 
-          svn status trunk | grep -q '^[ADMR!~]' || { echo "❌ No changes in trunk"; exit 1; }
+          # No -q here either, same reason as in "Update SVN tracking".
```
tyrann0us: **suggestion**

```suggestion
```

tyrann0us (changes requested):

.github/workflows/wordpress-org-release.yml
strangerkir: > Claude Code is very eager to comment as much as possible.

Yeah, true. But this one I kept intentionally to warn anybody who will be tempted to optimize it, like I was. `-q` is something one may want to add to speed things up: one finding is actually enough, so why continue searching? And the consequences of this are not obvious.

But I agree this sounds like an explanation why it is removed, not a warning. Maybe just rephrase it then? Something like "don't add `-q` to grep..." ?

.github/workflows/wordpress-org-release.yml
```diff
@@ -208,8 +208,10 @@ jobs:
           cd "${SVN_REPO_ROOT}/trunk"
           svn add . --force
           
-          #Only run if there's anything to delete
-          if svn status | grep -q '^!'; then
+          # Only run if there's anything to delete.
+          # No -q: it exits on the first match, SIGPIPE-ing svn status mid-write; under
+          # pipefail that non-zero exit wins over grep's, silently hiding real matches.
+          if svn status | grep '^!' >/dev/null; then
```
Biont: It could be that I am pedantic here, but with `-q` gone, are you not doing the same work twice now?

First, you do a full scan and if it produces ANY output, you do the full scan again to pipe it to `cut -c9- | while IFS= read -r file`

strangerkir: Good catch, thanks. Changed it so that we only grep status once, get the results, and then act on them.

This also makes completely redundant one of the comments Philipp was suggesting to remove.

.github/workflows/wordpress-org-release.yml
strangerkir: This comment is gone now. The code it was related to is gone now because of Moritz's finding.

.github/workflows/wordpress-org-release.yml
strangerkir: It was rewritten now as a warning to anyone who would like to optimize it by adding `-q` option for grep. Hopefully, this is better.

tyrann0us (approved):
Thanks!

PR #270, opened 2026-08-13
Ngrok parallel tunnel support for parallel tests

mishautkin:

---

Feature.

---

The ngrok used a single fixed domain, so parallel matrix jobs sharing that domain collided with ERR_NGROK_334. WordPress also wasn't aware that the forwarded requests were HTTPS.

---

By default the tunnel starts with no domain, so each job gets a random URL that is parallel-safe. An optionalc reserved NGROK_DOMAIN still can be used when set. NGROK_ENABLED allows to skip ngrok for jobs that don't require it. The forwarded port is taken from .wp-env.json testsPort, and an mu-plugin trusts X-Forwarded-Proto so WordPress treats tunnelled requests as HTTPS.

---

No.

---

## Security impact

Requires NGROK_AUTH_TOKEN as a secret; the tunnel steps are skipped when it's unset. While active, the tunnel exposes the test WordPress instance on a public URL for the duration of the run.

### DIFF
--- a/.github/workflows/test-playwright.yml
+++ b/.github/workflows/test-playwright.yml
@@ -72,6 +72,11 @@ on:
         default: ''
         required: false
         type: string
+      NGROK_ENABLED:
+        description: Whether to enable the ngrok tunnel integration.
+        default: true
+        required: false
+        type: boolean
       TESTRAIL_PLAN_ID:
         description: TestRail plan ID for reporting.
         default: ''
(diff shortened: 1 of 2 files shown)


.github/workflows/test-playwright.yml
```diff
@@ -247,10 +259,24 @@ jobs:
 
           echo "NGROK_URL=$NGROK_URL" >> "$GITHUB_ENV"
 
+      - name: Trust ngrok forwarded HTTPS scheme
+        env:
+          NGROK_AUTH_TOKEN: ${{ secrets.NGROK_AUTH_TOKEN }}
+        if: ${{ inputs.NGROK_ENABLED && env.NGROK_AUTH_TOKEN != '' }}
+        run: |
+          npx wp-env run tests-cli wp eval '
```
tyrann0us, line 267: "How many commands should we chain? - @mishautkin: "Yes!" 😆

tyrann0us (approved):

mishautkin:
Closed as testing proved it to be insufficient

PR #271, opened 2026-08-13
 Check Stable tag against SVN based on release mode

strangerkir:
### Please check if the PR fulfills these requirements

---

### What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)
Feature

---

### What is the current behavior? (You can also link to an open issue here)
Plugin versions from input, main plugin file, and readme must always match.

When creating a new release (`UPDATE_TRUNK_ONLY=false`), that version must also not already exist in SVN.
When only updating trunk (`UPDATE_TRUNK_ONLY=true`), no version-existence check is run at all.

---

### What is the new behavior (if this is a feature change)?
When `UPDATE_TRUNK_ONLY=true`, readme.txt's Stable tag is no longer required to equal PLUGIN_VERSION. Instead, it must reference a version that already exists as a published SVN tag.

This lets a not-yet-released version be synced to trunk for testing while Stable tag keeps pointing at the last real release, without WordPress.org serving trunk as "stable" in the meantime.

When `UPDATE_TRUNK_ONLY=false`, behavior is unchanged.

---

### Does this PR introduce a breaking change? (What changes might users need to make in their application due to this PR?)
Yes.

A caller with `UPDATE_TRUNK_ONLY=true`, whose readme.txt Stable tag currently equals `PLUGIN_VERSION` (i.e. a not-yet-tagged version) will now fail, since that tag doesn't exist yet. That's exactly the scenario this PR exists to fix. No production caller is affected today since none exist yet.

---

### Security impact
None.

---


.github/workflows/wordpress-org-release.yml
```diff
@@ -127,20 +130,30 @@ jobs:
...
+
+          # During a trunk-only sync, Stable tag is expected to still point to an already-
+          # published version, not PLUGIN_VERSION - it's checked against SVN's existing tags
+          # instead, once the SVN repository is checked out.
+          if [ "$UPDATE_TRUNK_ONLY" != "true" ]; then
+            CHECKED_AGAINST="$PLUGIN_FILE and readme.txt"
+            if [ "$VALIDATED_PLUGIN_VERSION" != "$README_VERSION" ]; then
+              echo "❌ Input version ($VALIDATED_PLUGIN_VERSION) does not match readme.txt Stable tag ($README_VERSION)"
+              MISMATCH=1
+            fi
```
Biont: **suggestion:**

This entire step is already quite long and contains a few conditionals.
Gathering the `README_VERSION` and this check here in a dedicated step gated on `if: ${{ inputs.UPDATE_TRUNK_ONLY }}` seems tidier to me.

I also do not think that the dynamic `CHECKED_AGAINST` log message is adding lots of value - especially if we gain structural clarity by extracting workflow steps.

Biont (changes requested):

.github/workflows/wordpress-org-release.yml
strangerkir: Agree on both. Updated.

Biont (approved):

